PR #8, opened 2024-04-25
Use `anyhow` for errors and `tracing` for logging

abhiaagarwal:
I was working on implementing Rustls, and was running into estetoric errors that were hard to debug. I realized this was because errors weren't getting properly propogated and instead were just being logged. This switches all unit-based error types to use `anyhow`, which is a Boxed dyn Error and allows upward propagation of errors.

This also uses tracing to log, instead of log. Tracing is the defacto standard for async code. I was running into some issue with backtraces, tracing solves this to a relatively ok degree.

The PR looks huge because I also `rust fmt` based on the the committed file, but it was mostly a simple find/replace and some manual tweaks.


hgaiser:
Hey :) and thanks for checking this out!

### `anyhow`

I'm familiar with `anyhow`, but I chose not to use it on purpose. The reason was that you are generally speaking best equipped to deal with an error in the place it occurs. Here is an example:

```rust
async fn run(self, stream: pulse_simple::Simple) -> Result<(), ()> {
	// Start recording.
	loop {
		// Allocate buffer for recording.
		let mut buffer = vec![0u8; 480];
			match stream.read(&mut buffer) {
			Ok(()) => {
				// Convert Vec<u8> to Vec<i16>.
				let samples = unsafe {
					Vec::from_raw_parts(
						buffer.as_ptr() as *mut i16,
						buffer.len() / std::mem::size_of::<i16>(),
						buffer.len() / std::mem::size_of::<i16>(),
					)
				};

				// Forget about our buffer, ownership has been transferred to samples.
				std::mem::forget(buffer);
					match self.audio_tx.send(samples).await {
					Ok(()) => {},
					Err(e) => {
						log::debug!("Received error while sending audio sample: {e}");
						log::info!("Closing audio capture because the receiving end was dropped.");
						return Err(());
					},
				}
			},
			Err(e) => {
				log::error!("Failed to read audio data: {}", e);
				return Err(());
			}
		}
	}
}
```

There's two ways this function can exit:

1. The receiving end was dropped. This is not a critical error, dropping the receiver is a "signal" that this thread should exit. Whether that is a good design choice is up for discussion of course ;) (here we don't know the reason the receiver was dropped, but then again it doesn't matter either, this thread should exit if there is no receiver), but that is how it is done at the moment.
2. Or it fails to read a sample from the audio device, which is an unexpected error.

In the first case you might want to log the event (`log::info`) and exit the function (this specific case probably doesn't even need to return a `Result` to be honest, since we are only returning `Err(())` and never `Ok(())` so this is probably a bad example, but I think for the purpose of discussion it is still valid).

The second case is more interesting because an unexpected error has occurred. In this case we want to log it as an error (`log::error`) and exit the function.

If we want to use `anyhow` here we have two options:

1. In both cases, return an `anyhow::Error` with the content describing the error. However by doing so we lose the information that the first event was a non-critical error, while the second one is a critical error. The function that handles the returned error doesn't know how to distinguish between the two.
2. Log first using `log::info` or `log::error` and then produce an empty `anyhow::Error`. This is basically what we do currently, in this case `anyhow` does not have any added value.

On an additional note: in an application I generally don't care what the exact error was. Otherwise I would have to use `thiserror` or some custom `Error` implementation. I only need to know if an error has occurred and either continue normal operation if it didn't, or exit early if it did.

I'm curious to hear what your opinion is on the added value of `anyhow`. I'm definitely open to other views, this was just my reasoning on why I implemented it like this :).

What was the issue you were trying to solve? You say the errors weren't propagated; do you mean the actual error messages, or `Err(())` itself? The latter should be propagated up if an unrecoverable error occurs, the former, at the moment, is not propagated up indeed.

### `tracing` vs `env_logger`

To be honest, I never used `tracing` before. What issues does it solve with async code? I haven't noticed any weird logging behavior with `env_logger` at least.

abhiaagarwal:
Hey! Thanks for the response.

I'll try to address both points here.

# Anyhow

In the case of the "error that's fine, we handled it", I believe it's better off to use `Ok(())` then as a return. In my case, I was interested primarily in backtraces, I don't want to see the backtrace of a function that correctly returned. I only tagged functions that were already returning an `Err(())` with `?`. There's also the relatively common `anyhow::Result<Option<Whatever>>>` (or `anyhow::Result<anyhow::Result<Whatever>>` if you want to get real funky. which represents the scenario you're describing, which is a "function that may not give us anything, but if it doesn't, we don't consider it an error, there are actual error cases". It's definitely more of a question of philosophy. I didn't change any logic with regards to "errors bubbling up", ie made functions early exit that didn't originally early exit.

Just to pick on your example, when you log an error, the caller loses any context about *why* the error occurred. This both on a technical level (no backtraces :() and a philosophical level. The caller can also perform error recovery if given strongly-typed errors with `thiserror` as you pointed out, but there's kind of a question of philosophy of "do we want the function to handle its error, or do we want the caller to handle a function erroring?" I think Rust code definitely prefers the latter, as functions should be "pure" and callers determine what to do with them. In the example, the error is logged, but it loses the capacity for graceful recovery. The caller can't inspect the logs — all it sees is an error of nothing. Returning a `Result<(), ()>` is basically returning a boolean (and I'm pretty sure that its size as well!) 

You're right that application-specific errors with `thiserror` are the "ideal" to go, and definitely can be used long-term, but this is application code, anyhow is well-used. It also adds additionally flexibility if you add any additional libraries with their own errors instead of spending time yak-shaving on deriving impls. While it's true that anyhow is essentially a `Box<dyn Error>`, its implementation is actually more like a `NonNull<Error>`, which means it uses only 8 bytes over 12 bytes (since the former is a "fat pointer". In my usecase, I ran into an issue with hyper + rustls, and since errors weren't being correctly propagated I couldn't view the backtrace with the common `RUST_BACKTRACE` env since the application in its current form essentially throws out the old error, converting it into a "logged string". The default `Display` impl of `std::Error` doesn't necessarily include a backtrace, anyhow adds the backtrace functionality (though I forgot to enable the feature in this PR, whoops!).

# Tracing

I removed env-logger + added tracing primarily because it's considered the default for async code, and it also has wide ecosystem support (pretty much the de-facto standard for async code). The upside is if you run `LOG=1`, you'll get tracing messages from other libraries propagated. I was running into an issue with Hyper (which honestly turned out to be an error on my side, as I was developing in debian and misconfigured a package). The additional upside is for async code (especially with Tokio), the default multithreaded runtime is work-stealing so a task may be resumed on another thread, meaning it's hard to track errors. Tracing adds the concept of `spans`, so a task keeps track of itself even if resumed on a different thread. I didn't implement it in this PR to keep it relatively contained, but all you have to do is add the macro `#[instrument]` to an async function to get the desired behavior. It also comes with the cool `tracing-appender` crate, which will auto log errors to a file without much configuration or `journald`.

abhiaagarwal:
Oh, I also forgot to mention, but there's likely to be a performance increase with this PR, though minimal (have not benchmarked :D). Logging to stdout/err is a relatively expensive operation (it's slower than a threading context switch for example on modern CPUs since it invokes a kernel privilege switch). Since writing to stdout is blocking, a function has an error, it has to block and write, then it returns — with `anyhow`, the error keeps going upwards until its ready to be consumed, where the context switch is acceptable since error-handling is on the cold path.

I will also note with regards to tracing, I haven't tested this but with the `#[instrument]` macro you can generate opentelementary traces which can be used to profile application performance

hgaiser:
> In the case of the "error that's fine, we handled it", I believe it's better off to use `Ok(())` then as a return.

Yeap, completely agree. If you happen to change that in this PR, feel free :).

> In my case, I was interested primarily in backtraces, I don't want to see the backtrace of a function that correctly returned.

Hmm maybe I'm misunderstanding something, or I'm about to learn something new :), but why would you see a backtrace of an `Error`? Wouldn't you only see a backtrace of a panic?

> Just to pick on your example, when you log an error, the caller loses any context about _why_ the error occurred. This both on a technical level (no backtraces :() and a philosophical level. The caller can also perform error recovery if given strongly-typed errors with `thiserror` as you pointed out, but there's kind of a question of philosophy of "do we want the function to handle its error, or do we want the caller to handle a function erroring?" I think Rust code definitely prefers the latter, as functions should be "pure" and callers determine what to do with them. In the example, the error is logged, but it loses the capacity for graceful recovery. The caller can't inspect the logs — all it sees is an error of nothing. Returning a `Result<(), ()>` is basically returning a boolean (and I'm pretty sure that its size as well!)

Wouldn't the same be true for `anyhow`? Since it can be any type of error, you're not going to gracefully recover based on the error that was returned and that also means you wouldn't really be inspecting the error from the calling function. I don't see how using `anyhow` makes a difference in terms of the context for the caller, versus returning `Result<_, ()>`. Both effectively provide no usable context.

Even if there is context (using `thiserror` or some other implementation), in most cases I can't see how it would be used in this case. Or at least, I can't think of a case where I would want to recover based on the error type. This is not the same for libraries though. If we're talking about a library I would prefer to return a specific error type, even if only for the reason that you don't know who is calling your functions and whether they might need to know. My point is that for applications I don't see an immediate need for having context of errors produced by the application itself.

I suppose I mostly don't see the added value of `anyhow`. I see a point for `thiserror`, it adds some utility making it easier to create custom errors, but `anyhow`, feature wise, seems the same as printing your error and returning `Err(())`.

> You're right that application-specific errors with `thiserror` are the "ideal" to go, and definitely can be used long-term, but this is application code, anyhow is well-used.

I know `anyhow` is used a lot, but that doesn't necessarily mean it's the right thing to do ;). If we don't need it, hey, that's (at least) one dependency less.

> It also adds additionally flexibility if you add any additional libraries with their own errors instead of spending time yak-shaving on deriving impls.

I can see how `anyhow` makes it easy to work with a lot of error types from different libraries, but printing the error and returning `Err(())` in essence does the same thing. The only requirement is that the error you receive implements `Display`, which is practically always.

> In my usecase, I ran into an issue with hyper + rustls, and since errors weren't being correctly propagated I couldn't view the backtrace with the common `RUST_BACKTRACE` env since the application in its current form essentially throws out the old error, converting it into a "logged string". The default `Display` impl of `std::Error` doesn't necessarily include a backtrace, anyhow adds the backtrace functionality (though I forgot to enable the feature in this PR, whoops!).

Hmm I believe this might be where I am missing something. I have a feeling this is mainly why you needed propagated errors? Can you tell me more about this? My only use of `RUST_BACKTRACE` has been to enable it if I'm working on a panic.

> # Tracing
> 
> I removed env-logger + added tracing primarily because it's considered the default for async code, and it also has wide ecosystem support (pretty much the de-facto standard for async code).

What makes it better suited for async code? I tried searching for this briefly but I didn't get a clear answer. I will dive deeper into `tracing`, but maybe you can help me out here?

`tracing`, with the `span` concept like you mentioned, does seem interesting for profiling where time is being spent, I can see that as a valuable addition. Having a tool to analyze how a single frame is processed could definitely be interesting :+1: .

ps. don't interpret my long replies as negative, I appreciate the discussion :).

abhiaagarwal:
No worries! I'm relatively seasoned rust dev, I would consider function code that logs itself to be un-idiomatic (see: this article by the author of *From Zero to Rust in Production*, which is a relatively-cited [book](https://www.lpalmieri.com/posts/error-handling-rust/#who-should-log-errors).

With regards to anyhow, the point is that it captures all priors errors in it with its internal backtrace. For example, let's say we have function A that calls function B, which calls function C, etc. Let's say that our final function, Z, returns an Err which may be propagated upwards. If Z were to log its error, we lose the context for why Z was even being called in the first place. From A's perspective, once the error gets to it, it just knows *something* has errored — there's no traceability between A and Z, even though A is (indirectly) responsible for Z's error. Y doesn't even need to necessarily react to the fact that Z errored or recovered from it, but Y probably wants to tell whoever called it that "hey, I called Z, it failed, here's what it told me, you figure out what you want to do with it". 

This is obviously a very contrived example, but anyhow adds that additional logging context that you were already using. Ie now, if there was an error `something bad has happened: error`, under the anyhow paradigm its now:

```
Something bad has happened

Caused by:
A: we messed up
B: we also messed up
C: we definitely messed up
```

`anyhow` stores its chains of errors internally. That's the upside of the library. 

I will also note that you can also directly downcast the error — so if `A` knew that `Z` returns only the `ZError` type, it can iterate on the `anyhow::Result.chain()` method once it arrives to them until it finds the correct error. Not particularly a substitute for proper error handling, but the functionality is there. 

I would argue primarily for application code this is important, because it helps users (and developers) have context into what errors occurred. For a less contrived example, once can imagine a read on a config file failing, under the current paradigm all the user would see in the logs would see the pretty-text for `EAGAIN` without any context for *why* the config file was being read in the first place. If there were a higher-level function doing the read on the config file, how would the user or developer know where its coming from? The context is immediately lost when logged.

I would also argue that it's important from an IO-perspective because you're basically buffering the error until someone wants to consume it, rather than just logging infinitely upwards. From a resources perspective, you're trading n writes to stdout for 1 write to stdout. Indeed, if we have multiple failures at the same time, it's impossible to associate which error is associated with another error, especially relevant in a concurrent/async environment. I haven't benchmarked but I'd imagine it can be substantial on the cold path. (fyi: I haven't submitted a PR for this but you might want to limit your multithreaded async runtime to use num::cpus() - 3 (audio, video, control threads) to avoid context switches).

And yes, for Rust 1.65+ errors do preserve their backtraces, which is only available with `RUST_BACKTRACE=1`. Which means that not only for panics, but for regular errors, you can view the exact chain of events that lead to it. See the [std::backtrace](https://doc.rust-lang.org/std/backtrace/) docs. It's only necessary if things truly go wrong, which happened in my case, but anyhow without the environmental variable provides the "pretty error chain".

Finally, with regards to tracing, yeah, the big upside is the `span` feature. Tokio "tasks" are just a green thread, and the tokio userspace scheduler can resume the task on any given thread. One can imagine a scenario. However, if we have some async task like this:

```rust
log::info!("hello world!");
let _ = some_future.await.map_err(|e| log::error!("we had an error part 1: {e}"));
log::info!("let's do some more work")
let _ = some_other_future.await.map_err(|e| log::error!("we had an error part 2: {e}"));
log::info!("I think we're done?")
```

Now imagine this task scheduled one million times. If this were logged to stdout, you don't have any real way to associate a task with where the error occurred. You can't use thread-local storage to assign an unique identifier because, well, tokio tasks are `Send` by default and we can't store local objects, so thread ids are useless. You could theoretically generate, say, an UUID per task, but that's expensive when scaling out. tracing::span (and the macro that automatically does this for you, `#[instrument]`) handles all this for you, generating a function whose tracing facades inside of it keep their context, even across different threads.

Also, from a pragmatic perspective, most async-based rust libraries using tracing, internally. Tokio uses it, hyper uses it, etc. It's impossible to read logs from `tokio` with `log`, but tracing has a compatibility layer with `log` that will automatically convert `log`s into traces.

hgaiser:
> No worries! I'm relatively seasoned rust dev, I would consider function code that logs itself to be un-idiomatic (see: this article by the author of _From Zero to Rust in Production_, which is a relatively-cited [book](https://www.lpalmieri.com/posts/error-handling-rust/#who-should-log-errors).

From my perspective the error is handled when it is logged. Returning `Err(())` is a signal up in the stack to let the caller know we didn't do what was asked.

> With regards to anyhow, the point is that it captures all priors errors in it with its internal backtrace. For example, let's say we have function A that calls function B, which calls function C, etc. Let's say that our final function, Z, returns an Err which may be propagated upwards. If Z were to log its error, we lose the context for why Z was even being called in the first place. From A's perspective, once the error gets to it, it just knows _something_ has errored — there's no traceability between A and Z, even though A is (indirectly) responsible for Z's error. Y doesn't even need to necessarily react to the fact that Z errored or recovered from it, but Y probably wants to tell whoever called it that "hey, I called Z, it failed, here's what it told me, you figure out what you want to do with it".

I think an important note here is that this only holds for `anyhow::Error` types. For example in the following case:

```rust
use anyhow::{bail, Result};

#[derive(Debug)]
struct SomeError {}
impl std::fmt::Display for SomeError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "SomeError")
    }
}
impl std::error::Error for SomeError {}

fn bar() -> Result<(), SomeError> {
    Err(SomeError {})
}

fn foo() -> Result<()> {
    bar()?;
    bail!("Foo");
}

fn main() -> Result<()> {
    foo()?;
    Ok(())
}
```

Since `SomeError` is not an `anyhow::Error`, it never shows up in the stacktrace:

```
Error: SomeError

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/hgaiser/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.82/src/error.rs:565:25
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/result.rs:1959:27
   2: foo::foo
             at ./src/main.rs:17:5
   3: foo::main
             at ./src/main.rs:22:5
   4: core::ops::function::FnOnce::call_once
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/ops/function.rs:250:5
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/sys_common/backtrace.rs:155:18
   6: std::rt::lang_start::{{closure}}
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/rt.rs:166:18
   7: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/ops/function.rs:284:13
   8: std::panicking::try::do_call
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:554:40
   9: std::panicking::try
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:518:19
  10: std::panic::catch_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panic.rs:142:14
  11: std::rt::lang_start_internal::{{closure}}
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/rt.rs:148:48
  12: std::panicking::try::do_call
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:554:40
  13: std::panicking::try
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panicking.rs:518:19
  14: std::panic::catch_unwind
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/panic.rs:142:14
  15: std::rt::lang_start_internal
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/rt.rs:148:20
  16: std::rt::lang_start
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/std/src/rt.rs:165:17
  17: main
  18: <unknown>
  19: __libc_start_main
  20: _start
```

Which makes it a bit strange to use in my opinion. The stacktrace appears complete, but it really only shows `anyhow::Error`s. Sure, if you use `anyhow` everywhere then it is complete, but it won't show errors from libraries or even from `std`.

> This is obviously a very contrived example, but anyhow adds that additional logging context that you were already using. Ie now, if there was an error `something bad has happened: error`, under the anyhow paradigm its now:
> 
> ```
> Something bad has happened
> 
> Caused by:
> A: we messed up
> B: we also messed up
> C: we definitely messed up
> ```

I see your point, but I don't think this is how I'd use it. By far most of the times it is fine to log an error in the place where it occurs. The required information is already there. There's definitely examples where this is not true, but they are quite rare. One that comes to mind is file parsing. A parsing function might know what went wrong during parsing, but not what file(path) was being parsed. This doesn't require a crate to solve though. Simply returning an (actual) error in the parsing function would suffice.

> I would argue primarily for application code this is important, because it helps users (and developers) have context into what errors occurred.

I disagree here. I think logging should be simple and straightforward to interpret by a user. If a developer wants more information then it's possible to extract more from, for example, the `tracing` events. You wouldn't want a user to see a full stack backtrace for instance.

> I would also argue that it's important from an IO-perspective because you're basically buffering the error until someone wants to consume it, rather than just logging infinitely upwards. From a resources perspective, you're trading n writes to stdout for 1 write to stdout. Indeed, if we have multiple failures at the same time, it's impossible to associate which error is associated with another error, especially relevant in a concurrent/async environment.

Hmm I disagree here too. We're not logging infinitely upwards. We log an error, return `Err(())` and in the calling function we mostly (always?) return without logging too.

> (fyi: I haven't submitted a PR for this but you might want to limit your multithreaded async runtime to use num::cpus() - 3 (audio, video, control threads) to avoid context switches).

Interesting point, hadn't thought of that!

> And yes, for Rust 1.65+ errors do preserve their backtraces, which is only available with `RUST_BACKTRACE=1`. Which means that not only for panics, but for regular errors, you can view the exact chain of events that lead to it. See the [std::backtrace](https://doc.rust-lang.org/std/backtrace/) docs. It's only necessary if things truly go wrong, which happened in my case, but anyhow without the environmental variable provides the "pretty error chain".

This is only true for `anyhow::Error`'s, same as the point above.

> Finally, with regards to tracing, yeah, the big upside is the `span` feature.

I'm sold on `tracing`, I'd like to give it a go. If you want you can submit a separate PR for `tracing` so we can get that merged sooner?

hgaiser:
Regarding the backtrace feature from `anyhow`, you only get this if you're returning the error from the main function. A lot of the processing happens in tasks, of which the handles have been discarded. Meaning in many cases there's no more backtrace to print.

I'm leaning towards not going for `anyhow`, but like I said in my previous post, I'd like to add tracing :)

hgaiser:
Let me know if you're interested in making a `tracing` PR, otherwise I'll pick it up sometime :+1:

hgaiser:
I added the `tracing` parts in https://github.com/hgaiser/moonshine/pull/9 . Unfortunately I couldn't cherry-pick from this PR, but I wanted to thank you for the discussions and for informing me about `tracing`, looks cool :+1: 

I will close this PR in favor of https://github.com/hgaiser/moonshine/pull/9 . Regarding `anyhow`: I don't see the added value for that currently. You're welcome to change my mind :).

PR #19, opened 2026-02-13
Enable AV1 encoding support

porkloin:
:wave: this adds AV1 support to moonshine. It depends on https://github.com/hgaiser/pixelforge/pull/1 - once that PR is merged I will need to update the cargo.toml to reflect the new pixelforge version.

You can test it by building moonshine locally and pointing the pixelforge dep in cargo.toml at a local binary build from the pixelforge pull referenced above.


hgaiser (commented):
Does this work without enabling [this](https://github.com/hgaiser/moonshine/blob/main/src/webserver/mod.rs#L434-L437)? Maybe enabling that will make the client request AV1 when it is set to Automatic.

Could you enable those flags?

PR #20, opened 2026-02-13
Fix Moonlight discovery, make uniqueid optional

porkloin:
When running moonshine, I had to manually input the IP address of my host machine. The logs in the systemd unit showed a bunch of traffic complaining about `uniqueid` being null.

I compared this to how sunshine does it and it looks like discovery traffic sends probing requests that don't contain uniqueid. This implements some tolerance for traffic that doesn't contain `uniqueid` so that discovery requests work.


hgaiser:
Thanks! Sorry for not getting to this earlier :)

hgaiser:
Ah I also see some pipewire code in this PR. Could you remove those parts, since they're no longer used?

porkloin:
Yup I'll rebase this one with only the discovery bits, might be a couple days til I can get to it though

PR #21, opened 2026-02-14
Graceful session shutdown, two-tier stream lifecycle, Steam integration

porkloin:
This adds a bunch of general improvements to the lifecycle for launching streams and especially for launching steam games with the steam integration.

Previously, every stream that I ran would start and fail to initialize a control stream. After a few seconds moonlight client (iOS) would return to the "app overview" and the application I had selected would be shown with an active but disconnected stream, which I could successfully reconnect to.

This fixes that by establishing a better async connection pattern for stream initialization, and prioritizes getting a session within the timeout impose by clients like moonlight.

This also delegates some extra graceful shutdown logic to the moonshine daemon, which has a much longer available window to allow graceful shutdowns.

Lastly, when streams initiated with the steam integration are shut down, this spawns a new steam on the host machine post-disconnect to ensure we aren't leaving the host in a state where steam is killed when it was running on the host before the stream began.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,7 +49,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.16",
+ "getrandom 0.2.17",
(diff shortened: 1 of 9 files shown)


porkloin:
Converting this to draft for now til #19 and https://github.com/hgaiser/pixelforge/pull/1 are resolved

PR #29, opened 2026-03-12
Allow for multiple controllers

antlilja:
Moonshine does currently not allow for multiple controllers at the same time because the controller id is duplicated for every controller. This PR fixes this by basing the id on the gamepad index.


hgaiser:
Strange that I didn't have this issue, but I don't see a problem with merging this if it fixes your issue. Thanks for the PR!

PR #37, opened 2026-03-18
Add desktop application scanner for .desktop launchers

antomanc:
Add a desktop application scanner that discovers applications from configured directories containing .desktop files.
Configurable like the Steam Library scanner.
Tests and build are passing and it works pretty well for Lutris .desktop shortcuts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,17 +44,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.7.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


src/app_scanner/desktop.rs
```diff
@@ -0,0 +1,848 @@
...
+			Ok(metadata) => metadata.permissions().mode() & 0o111 != 0,
+			Err(_) => false,
+		}
+	}
+
+	#[cfg(not(unix))]
+	{
+		true
+	}
+}
```
hgaiser: This is a bit redundant for us since we'll always run on unix.

Issue #50, opened 2026-03-30
Wayland native apps bind to the wrong Wayland socket after the first streaming session


me-teorick:
When a Moonshine streaming session stops, Wayland sockets are not cleaned up by Moonshine's embedded compositor (Smithay). With several sessions, the sockets keep on incrementing (`wayland-1`, `wayland-2`, `wayland-3`, etc.), and are only cleaned up upon reboot.

At the same time, when Moonshine launches an application via `systemd-run`, it explicitly removes the `WAYLAND_DISPLAY` environment variable. As per the comments in the source code, this is intentional, to force games to use X11 via XWayland. However, Wayland-native applications (e.g., Waydroid) use this variable to determine which Wayland socket to bind to, and default to "wayland-1" when the variable is not set. Because "wayland-1" is only used during the first streaming session and it becomes stale when that session ends, the subsequent streaming sessions show a blank screen because the Wayland-native app renders its UI to a stale socket.

A workaround is to wrap the launch of the Wayland-native app inside a script that sets `WAYLAND_DISPLAY` to the most recently created Wayland socket:

```bash
export XDG_RUNTIME_DIR=/run/user/$(id -u)
# Find the most recently created wayland-* socket
export WAYLAND_DISPLAY=$(ls -t $XDG_RUNTIME_DIR/wayland-* | grep -v '\.lock$' | head -1 | xargs basename || echo "wayland-1")
```

While this works in practice, it is not ideal.

hgaiser:
I expect this will be fixed by https://github.com/hgaiser/moonshine/pull/68

hgaiser:
I will assume this is fixed by #68 . If the issue persists, feel free to re-open the issue :).

me-teorick:
> I will assume this is fixed by [#68](https://github.com/hgaiser/moonshine/pull/68) . If the issue persists, feel free to re-open the issue :).

I finally got to test this, I can confirm it is working perfectly. Thanks!

PR #51, opened 2026-03-30
fix: enable Wayland-native apps to bind to the embedded compositor's socket

me-teorick:
A possible fix to #50

I do not know much about Wayland, so I went with the assumption that Smithay not cleaning up sockets is a best practice to avoid racing conditions and so on. Under that assumption, it is best to leave current code mostly unchanged and simply add a `wayland_native`  boolean field to  `ApplicationConfig`  that tells whether the application is Wayland-native and thus controls whether  `WAYLAND_DISPLAY`  should be set or unset when launching that specific application. Defaults to  `false`  for backward compatibility. The rest of the code changes merely follow from that decision.

Config example:
```toml
[[application]]
title = "Waydroid"
command = ["/usr/bin/waydroid", "show-full-ui"]
wayland_native = true
```

Scanned applications have `wayland_native = false` for now, since this feature is mostly used to populate entries from the installed Steam Library. May need to rethink it if other use cases come up.

### DIFF
--- a/src/app_scanner/desktop.rs
+++ b/src/app_scanner/desktop.rs
@@ -94,7 +94,12 @@ fn parse_desktop_application(
 
 	let boxart = icon_resolver.resolve(entry.icon().map(str::trim).filter(|icon| !icon.is_empty()), path);
 
-	Ok(Some(ApplicationConfig { title, boxart, command }))
+	Ok(Some(ApplicationConfig {
+		title,
+		boxart,
+		command,
+		wayland_native: false,
+	}))
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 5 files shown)


me-teorick:
Okay. But regardless of whether Smithay should clean up its sockets, scenarios in which several valid wayland sockets are being used at any given moment are not hard to imagine (even if not so common). Wayland-native apps need WAYLAND_DISPLAY to know which socket they should bind to, otherwise they will bind to a default one that only by chance will be the one used by Moonshine.

me-teorick:
Then it seems that this PR code is doing its job and you hit the next problem... It may be related to your GPU(s). To be 100% sure, don't merge this PR and use the workaround of wrapping the Wayland-native app call in a script that sets the WAYLAND_DISPLAY variable as mentioned in issue #50 . Both should be functionally equivalent, but the script is easier to debug.

Issue #52, opened 2026-03-31
Release 0.9.0 breaks compatibility with LG TV moonlight clients


me-teorick:
I am using [Moonlight-TV](https://github.com/mariotaku/moonlight-tv) and its [Aurora fork](https://github.com/GuiDev1994/moonlight-tv) to connect to `moonshine` from my LG TV. With the new security-focused enhancements introduced in release 0.9.0, these clients stopped being able to talk to `moonshine` due to `TLS handshake failed: invalid peer certificate: Other(OtherError(UnsupportedCertVersion))`. This is because they seem to generate certificates of type **`X.509 v2`** instead of **`X.509 v3`**.

The reason why these clients use `X.509 v2` is unclear to me, but it is likely that during the OpenSSL to MbedTLS migration (commit `4f674cfa` on 2021-03-07), the `moonlight-tv` developer incorrectly translated the certificate version constant:

**Original OpenSSL code:**
```c
X509_set_version(x, 2);  // OpenSSL uses 0-indexed: 0=v1, 1=v2, 2=v3
```

**Broken MbedTLS code:**
```c
mbedtls_x509write_crt_set_version(crt, MBEDTLS_X509_CRT_VERSION_2);  // =1, which is X.509v2
```

**Correct MbedTLS code should be:**
```c
mbedtls_x509write_crt_set_version(crt, MBEDTLS_X509_CRT_VERSION_3);  // =2, which is X.509v3
```

This has not been much of an issue to anyone until now because most servers are lenient - Sunshine, GeForce Experience, and NVIDIA's GameStream don't strictly validate client certificate versions. Sunshine in particular seems to accept any certificate during TLS handshake but then performs validation of some of the certificate credentials in the application layer.

I will patch `moonlight-tv` for my personal use and eventually submit a PR over there. In this project though, I wanted to ask: is strict security more important than broad client compatibility? If the latter, I can submit a PR that either makes strict security optional via a config file param or implements Sunshine-style credential handling (or both). If security is most important, then feel free to close this as "won't fix".

PR #55, opened 2026-03-31
feat: make requiring/verifying client certs optional

me-teorick:
A possible fix to #52 

The rationale was to make the security enhancements introduced in moonshine v0.9.0 optional via configuration parameters. 

Added two configuration parameters to `WebserverConfig`, which default to `true`:
 
```toml
[webserver]
# Whether to require client certificates for HTTPS endpoints.
# When enabled (the default), HTTPS endpoints require a valid client certificate 
# from a paired client. Disable this for compatibility with clients that do not 
# support mTLS.
require_client_cert = true
 
# Whether to enforce strict X.509 v3 certificate verification.
# When enabled (the default), only X.509 v3 certificates are accepted. Disable 
# this to accept older X.509 v2 certificates (e.g., from moonlight-tv). This 
# bypasses TLS-level signature verification for client certificates.
strict_cert_verification = true
```

Perhaps both parameters should default to `false`? 
(so that first time users get a "smooth" experience with any client, while those who need the extra security will surely be able to find these options in the docs)

I tested this on my LG TV with `require_client_cert = true` and `strict_cert_verification = false`, and the Aurora client was able to connect to moonshine again.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -104,6 +104,22 @@ pub struct WebserverConfig {
 	#[serde(default = "default_true")]
 	pub enable_pairing: bool,
 
+	/// Whether to require client certificates for HTTPS endpoints.
+	///
+	/// When enabled (the default), HTTPS endpoints (/launch, /resume, /cancel, /applist, /appasset)
+	/// require a valid client certificate from a paired client. Disable this for compatibility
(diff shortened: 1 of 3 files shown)


hgaiser:
This has been superseded by #66

PR #70, opened 2026-04-16
fix: handling multiple steam libraries in steam app scanner

reeves-48777:
I had an issue where moonshine could not retreive my game list from steam, and I found out that it was because they were installed on another drive.

Here is what I did :

- Added a new struct that represents a simplified LibraryFolder
- Added a parser function (called parse_vdf) that generates a vector of LibraryFolder
- Changed the get_game_name function to use the steamapps_dir folder given in argument
- Changed scan_steam_applications function to use parse_vdf

### DIFF
--- a/src/app_scanner/steam.rs
+++ b/src/app_scanner/steam.rs
@@ -7,6 +7,75 @@ use walkdir::WalkDir;
 
 use crate::config::{ApplicationConfig, SteamApplicationScannerConfig};
 
+struct LibraryFolder {
+	path: PathBuf,
+	apps: Vec<u32>,
+}
+
+fn parse_vdf(vdf_content: &str) -> Vec<LibraryFolder> {
+	let library_folder_depth = 2;
+	let apps_depth = 3;
+	let path_pattern = "\"path\"";
+	let mut library_folders = Vec::new();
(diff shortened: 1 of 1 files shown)


hgaiser:
Thanks for the PR! I'm currently on vacation so I'll check this in a few days.

Does this issue occur when multiple game dirs are configured in Steam? Or just from having the game dir on another drive?

reeves-48777:
Hi! Sorry for the late reply!

On my setup I have two game dirs, each on a different disk. However, I think it would occur in both situations.

The `libraryfolders.vdf` file looks like this when there are two game dirs:
```vdf
"libraryfolders"
{
    "0"
    {
        "path" "~/.local/share/Steam/steamapps/"
        ...
        "apps"
        {
            "Game Id" ...
            ...
        }
    }

    "x"
    {
        "path" "/path/to/folder"
        ...
        "apps"
        { ... }
    }
}
```

This block repeats for each game directories configured.
The current parser search for the first "apps" key occurrence and then skips all the subsequent "apps" keys within the file, missing games installed on other drives or directories.

I hope it helps :)

Thank you for your answer and taking the Time for checking my PR!

hgaiser:
Thanks!

PR #71, opened 2026-04-19
fix: swap chain detection for games that don't use vkSetHdrMetadataEXT

porkloin:
Fixes HDR not activating for games that signal HDR via swapchain colorspace instead of vkSetHdrMetadataEXT. For me, this was happening in MH Wilds, moonshine would always just start the stream and stay in SDR mode regardless of how the game HDR setting was set. It's a bit of an odd case since it doesn't have an explicit toggle for "hdr on" it's either "automatic" or "disabled". I think unfortunately different game engines are all over the place with how they signal HDR :(

also rolled in a few related things from the same investigation that cut host processing latency for games that use vkd3d-proton (RE engine games primarily) by a significant amount:

- Direct DMA-BUF scanout. The compositor was incorrectly taking the override surface into its own buffer which resulted in resource contention with the game rendering buffer if the GPU is fully saturated. 
- Caching for imported DMA-BUFs by wl_buffer ObjectId. Some games using vkd3d-proton (DXVK's d3d12 layer) create a fresh wl_buffer per frame instead of reusing a pool, so the cache keying strategy always missed, resulting in expensive operations for every frame trashing the host rendering performance. This doesn't seem to happen on D3D11, opengl, or vulkan games at all. Again, I was able to cause this with MH Wilds and Pragmata, but as far as I know any RE Engine title from within the last ~5 years should be impacted.
- Map XBGR2101010 / XRGB2101010 fourcc to10-bit input format. With direct scanout, this bug surfaces at 10-bit override-surface frames getting treated as 8-bit and producing classic "rainbow" corruption in HDR. caught while debugging the HDR detect issue.

This should be something that you can recreate by A/B testing streaming an RE engine game. Before and host rendering performance should be significantly improved, and HDR detection should also work properly. You may need to crank game settings to max and turn off frame gen/DLSS/FSR to get a significant enough GPU load to see the performance issues, depending on your GPU.

### DIFF
--- a/moonshine-wsi/src/swapchain.rs
+++ b/moonshine-wsi/src/swapchain.rs
@@ -47,7 +47,7 @@ pub unsafe extern "C" fn create_swapchain(
 	// compositor signals HDR support; otherwise pass the app's create-info
 	// through unchanged so the ICD can handle its own color management.
 	let layer_hdr_active = instance_key
-		.and_then(|ik| get_wayland_connection(ik))
+		.and_then(get_wayland_connection)
 		.map(|arc| arc.force_lock().caps.hdr_supported)
 		.unwrap_or(false);
 	let need_remap = layer_hdr_active && app_color_space != ash::vk::ColorSpaceKHR::SRGB_NONLINEAR;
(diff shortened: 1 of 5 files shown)


hgaiser:
Thanks! Looks good to me.

hgaiser:
I did have to adjust this slightly in https://github.com/hgaiser/moonshine/commit/bd51fbd8bf547effb039eb00be12d0bcc90ec4cc because the cursor was hidden in some games. With that change, when a cursor is shown it will not use direct scanout. I'm not sure if this is the correct approach 🤷

PR #72, opened 2026-04-19
fix: infer virtual display dimensions from requested resolution

porkloin:
Some games seem to be upset if the display dimensions are 0x0, this should fix that


hgaiser:
Thanks!

PR #73, opened 2026-04-19
fix: Stop existing Steam before launching streamed application

porkloin:
I keep running into a problem where if the host already has steam running, moonshine fails to start a remote streaming session. This kills any existing steam instance on the host before starting the moonshine stream.

### DIFF
--- a/src/session/mod.rs
+++ b/src/session/mod.rs
@@ -1,5 +1,6 @@
 use std::process::Command;
 use std::process::{Child, Stdio};
+use std::time::Duration;
 
 use async_shutdown::ShutdownManager;
 use manager::SessionShutdownReason;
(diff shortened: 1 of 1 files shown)


porkloin:
I'm realizing now that the way I run the systemd unit (--user rather than system level) means it's associated with my actual login session, so this might not be necessary at all for the intended install pattern :thinking:

hgaiser:
I mean the intended install pattern is either with a running desktop environment, or without. However I feel like this is an issue that could and should be resolved with pre_command settings for Steam applications, in the config file.

porkloin:
> I mean the intended install pattern is either with a running desktop environment, or without. However I feel like this is an issue that could and should be resolved with pre_command settings for Steam applications, in the config file.

Fair, I'm happy to close this in favor of just adding a pre_command, I didn't think of that!

PR #74, opened 2026-04-20
feat: add pre/post cmds

porkloin:
- Adds the ability for arbitrary `pre_command` and `post_command` that can be run before/after and defined for each different application types in `config.toml`
- Example: `pre_command = ["pkill", "-x", "steam"] ` to kill existing steam instances before starting steam-based streaming targets


hgaiser:
I changed the code slightly to allow multiple pre/post commands to be set. After CI finished I'll merge this, thanks 👍

PR #76, opened 2026-04-23
Add supersampling support

dylanfrankcom:
Renders the Wayland compositor at 2× the client-requested resolution for resolutions listed in `supersampled_resolutions`, then GPU-downscales to encode resolution before the color converter. Improves anti-aliasing on low-resolution clients at the cost of higher GPU load.

For example, running the Wayland compositor at 2560x1600 when being streamed to the Steam Deck at 1280×800. This improves anti-aliasing while keeping required streaming bandwidth low, and gives FSR/DLSS more resolution to work with instead of a client's low native resolution.

Configured in `config.toml` as e.g. `supersampled_resolutions = [[1280, 800], [1920, 1080]]`.

Tested with multiple resolutions against multiple titles running under the default XWayland and native Wayland via GE-Proton.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -226,6 +226,12 @@ pub struct VideoStreamConfig {
 	/// the messages can be very noisy.
 	#[serde(default = "default_false")]
 	pub log_frame_spikes: bool,
+
+	/// List of client resolutions (width, height) that will be rendered at 2× and
+	/// downscaled before encoding. Empty list disables supersampling entirely.
+	/// Example: `supersampled_resolutions = [[1280, 800], [1920, 1080]]`
+	#[serde(default)]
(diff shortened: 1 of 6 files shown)


hgaiser:
Thanks! Would be nice if Moonlight allows setting this in their options instead of defining resolutions that will be supersampled.. maybe something for another day.

Could you have a look at the remarks from Copilot? I am not that familiar with Vulkan so I can't really judge that part well. The rest looks good to me :+1:.

hgaiser:
I'm closing this in favor of https://github.com/hgaiser/moonshine/blob/main/TIPS.md#use-gamescope-with-the-clients-resolution

I'm not opposed to the idea of having it inside moonshine, but I don't see the need for it at the moment. I think for now it's better to use gamescope if you want to do super sampling.

PR #78, opened 2026-04-27
feat: pass convert semaphore through to encoder.

porkloin:
This is a companion to [this pixelforge PR](https://github.com/hgaiser/pixelforge/pull/10). After that is merged, pixelforge's `Encoder::encode()` now takes an optional GPU wait semaphore so convert and encode can overlap on the GPU instead of CPU-blocking between them. this wires the semaphore returned by `converter.convert()` through to the encoder submit so the wait happens GPU-side. That PR has all the details on how and why this works, but tl;dr is it improves convert+encode by parallelizing some stuff

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2756,7 +2756,7 @@ dependencies = [
 [[package]]
 name = "pixelforge"
 version = "0.3.0"
-source = "git+https://github.com/hgaiser/pixelforge?rev=fc60c32#fc60c32dcc7684467bd0da3ab6f0cf56975a235d"
+source = "git+https://github.com/porkloin/pixelforge?rev=e1fb696#e1fb696f29af54d3bf0b4550402f9473b92f43fc"
 dependencies = [
  "ash",
  "shaderc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2756,7 +2756,7 @@ dependencies = [
 [[package]]
 name = "pixelforge"
 version = "0.3.0"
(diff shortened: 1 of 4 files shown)


hgaiser:
Any idea what the speedup is with this approach? I'd be interested to see the benchmark results :)

In my earlier testing, color conversion took a few hundred microseconds so I'd imagine the gain is not very large.

porkloin:
> Any idea what the speedup is with this approach? I'd be interested to see the benchmark results :)
> 
> 
> 
> In my earlier testing, color conversion took a few hundred microseconds so I'd imagine the gain is not very large. 


Totally, I put some metrics on the  PR for the pixelforge side: https://github.com/hgaiser/pixelforge/pull/10

TL;DR is yes it's not a huge gain for convert since as you say it's microseconds, but this unlocks doing parallel/non blocking processing which the real win and can provide extra headroom especially when throughput is completely saturated

hgaiser:
Wow, that `VK_VALVE_video_encode_rgb_conversion` extensions looks nice! How are the benchmarks looking?

porkloin:
> Wow, that `VK_VALVE_video_encode_rgb_conversion` extensions looks nice! How are the benchmarks looking?

Yeah! It makes convert basically free, but it requires a change on both sides. Kinda like how I did with https://github.com/hgaiser/pixelforge/pull/10 being closed in favor of https://github.com/hgaiser/pixelforge/pull/12 I'll do the same here. There will need to be a pixelforge side PR for vk valve also. But tl;dr it makes convert "free" which might open the door for some higher framerates than moonlight servers have traditionally been able to support.

For now, closing this to drop the semaphore stuff, but the vk  valve stuff will come back in its own dedicated PR

PR #84, opened 2026-05-10
Refactor app launch flow to show errors more clearly.

hgaiser:
Refactor how an application is launched so that errors can be sent to the moonlight client.


jruelle23:
Can confirm it prints the message on my end when reproducing my initial error:
image

hgaiser:
Great! Thanks for testing. It's by no means a very user friendly way to present this specific error, but it's better than a generic error message 🤷 

This change seems to have broken HDR support. I'm working to fix that before merging.

PR #88, opened 2026-05-13
Dockerfile

clemperorpenguin:
Hi! I thought this would be especially useful in a container with Steam, Lutris, and ES-DE. I've set that up here, it works well with my limited testing (AMD, Fedora 44, in Podman).

Currently, it installs the stable moonshine and ES-DE from the AUR, this can be changed to git if that is preferred. Further, we could save significant space by making two Dockerfiles, as the NVIDIA one has several extra dependencies, but I figured it would be easier/best to target all platforms in one.

The other unusual thing I did was add a systemd-run shim to make it unnecessary to have systemd running in the container, which can be cumbersome to manage.

Let me know if anyone has any feedback, or even if this isn't needed at this time, I just thought I'd share since I'll likely keep using this.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,15 @@
+name: Docker Image CI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Build the Docker image
+      run: docker build . --file Dockerfile --tag moonshine:${{ github.event.release.tag_name }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM archlinux:base
 
 # 1. Configure pacman
-RUN sed -i '/^#\[multilib\]/s/^#//' /etc/pacman.conf && \
-    sed -i '/^#Include = \/etc\/pacman.d\/mirrorlist/s/^#//' /etc/pacman.conf && \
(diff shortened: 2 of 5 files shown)


hgaiser:
Hmmm I'm not necessarily against a Dockerfile, but I'm wondering if at that point you're maybe better off using https://github.com/games-on-whales/wolf , which is basically a Moonlight server in a container.

Could you explain why you'd use this (moonshine + Dockerfile) over wolf?

clemperorpenguin:
With all due respect to the authors of wolf, which I've used before and enjoyed using, I found it really difficult to manage, a little unstable, and I found the per-client library architecture (not sure how else to say that lol)/workarounds to be frustrating. tldr I can look again, but they seemed really focused on the use case of running multiple streaming and gaming instances on one PC instead of just a basic gaming container that you can point at an existing lutris and steam library. - iirc I had some difficulty setting up an existing library because it sets them up for you per-client.

All of that said, I can take another look at wolf and how it is working now, it's been a while since I tried it or read about it much.

hgaiser:
I will not merge this for now. I'm not necessarily against a Docker container, but I don't see the problem that it solves.

It could be a way to deploy moonshine easily, but then I'd prefer to investigate alternatives (like a Flatpak package).

It could serve as a way to isolate sessions, but then I'd rather suggest using multiple user accounts and run as `systemctl start moonshine@user1`, `systemctl start moonshine@user2` etc. (although that is a bit more work to setup).

To summarize: thanks for checking this out and putting in the effort to make it work in a container! But for now I will close this.

PR #92, opened 2026-05-19
feat(release): add binary and vulkan layer to release assets

porkloin:
- Adds prebuilt binary to release assets
- Adds vulkan layer to release assets
- Moves CI to run on ubuntu 22.04 (better compatibility for unlinked deps on prebuilt binaries)
- Source builds shaderc during build to static link, since this currently the most likely dependency that will be inconsistent across host systems

@hgaiser this should be a first step to make distribution outside of AUR easier, that way whatever system gets used, there's an easy way to fetch binary and vulkan layer from the latest release rather than needing to build from source every time

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,9 +3,11 @@ name: Release
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
+  DEBIAN_FRONTEND: noninteractive
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


porkloin:
Putting this into draft for now, need to think a bit about build time and run time deps a bit more and figure out what might need to be source-built and linked during CI, but I think there's a road forward to get a single usable moonshine binary for the vast majority of users.

hgaiser:
Have you had any progress on this? I was checking the linked libraries for the moonshine binary and managed to remove some by making the shader a binary blob (instead of compiling it on the fly).

https://github.com/hgaiser/moonshine/commit/a9860e699472feee5108455358608a969f9d9f81

porkloin:
I'm hoping maybe I can take a look at this in the next week or so :crossed_fingers:

porkloin:
@hgaiser just rebased and updated, shaderc is no longer required to be source built and static linked. That means the main thing left in the changes here is moving to the ubuntu 22.04 runner which should ensure that systems running the binary that have glibc 2.34 or newer which should cover Fedora 35+, Debian 12+, Ubuntu 22+ etc. Otherwise if it was built with the arch runner we'd end up excluding any distros with lower than 2.43

hgaiser:
Thanks again! I changed the implementation a bit in #130 to also include .deb and .rpm packages and I used ubuntu 24.04 since 22.04 is EOL within a year already.

PR #96, opened 2026-05-31
Fix PulseAudio stream deadlock causing audio to cut out permanently

Arecsu:
When streaming games through Wine/Proton (via Lutris, etc.), audio would intermittently cut out entirely or cycle between "audio burst → silence → audio burst" sounding robotic/laggy. The issue was a deadlock in the PulseServer's Request mechanism.

The server tracks pending byte Requests via `requested_bytes`. If a client pauses or slows down its writes (which winepulse does during loading screens, menu transitions, etc.), `requested_bytes` accumulates past `target_length`. Once that happens, `bytes_needed` becomes `0` and the server stops sending Request commands entirely, even after the buffer drains to empty. The client never gets asked for data again, so audio stays silent forever (or until the next reconnection cycle).

The fix resets `requested_bytes` to 0 whenever the buffer falls below `minimum_request_length`, ensuring a fresh Request is always dispatched when the client runs low on data.

### DIFF
--- a/src/session/stream/audio/pulse_server/mod.rs
+++ b/src/session/stream/audio/pulse_server/mod.rs
@@ -478,18 +478,26 @@ impl PulseServer {
 					}
 				}
 
-				let bytes_needed = (stream.buffer_attr.target_length as usize)
-					.saturating_sub(stream.buffer.len_bytes() + stream.requested_bytes);
+				let min_req = stream.buffer_attr.minimum_request_length as usize;
+				let buf_len = stream.buffer.len_bytes();
+				let target = stream.buffer_attr.target_length as usize;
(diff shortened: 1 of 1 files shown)


hgaiser:
Thanks for the PR! I looked into this and opted for a slightly different approach in https://github.com/hgaiser/moonshine/commit/a194dcc4f852c03a82fe4b8f4f251406734d0127 in the end. Let me know if your issue is resolved with that commit 👍

Arecsu:
Nice! I'll be testing it sometime in the span of this week and see how it goes, if I find something wrong then I'll let you know. For now, bet you it is solved, we'll see!

Arecsu:
In the span of the next minutes I should've said lmao. Working awesome!

hgaiser:
Haha cool, thanks for checking!

PR #97, opened 2026-05-31
Scroll: normalizing scroll speed, particularly for macOS clients + using the discrete wayland scroll protocol

Arecsu:
When connecting from a macOS client (trackpad or mouse), every little scroll tick moved way too far, like 12 steps instead of 1. Sunshine felt much better. Inspecting the codebase of both, Sunshine is indeed doing something special about it, but given moonshine uses wayland directly, we need a different strategy.

My own fix introduced a bug where scrolling one direction then switching (e.g. down 4× then up) would make the first scroll in the new direction do nothing. That's fixed too in the same commit.

Now scrolling feels smooth, controlled, and direction changes work on the first scroll. Maybe even better than Sunshine?

I have no other client devices to test this further though. Only macOS. It should work properly on other clients, I guess.

---

Added a second fix: scroll events were not passed correctly to games/apps running under gamescope with `--backend wayland`.

Scroll events used the basic `wl_pointer.axis` event. Some Wayland clients (like the previously mentioned gamescope) ignore the basic axis event and rely on the newer discrete/high-resolution scroll protocol events. So this second commit adds `axis_value120` and `axis_discrete` alongside the basic event so scrolls reach every client regardless of which protocol version they use.

### DIFF
--- a/src/session/compositor/input.rs
+++ b/src/session/compositor/input.rs
@@ -13,6 +13,17 @@ use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerCons
 
 use super::state::MoonshineCompositor;
 
+/// Wheel tick size matching Windows WHEEL_DELTA convention.
+/// Moonlight clients send scroll deltas in high-res units where
+/// 120 = one scroll notch. These are accumulated until reaching
+/// this threshold before emitting a scroll tick.
+const WHEEL_DELTA: i32 = 120;
+
+/// Wayland axis value for one scroll notch.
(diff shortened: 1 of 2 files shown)


hgaiser:
Thank you for looking into this! I always felt the scrolling was a bit too much :). I looked at your suggestion but I think it just needed proper conversion from `WHEEL_DELTA` to degrees. I fixed this in https://github.com/hgaiser/moonshine/commit/8646fbcca481c77c01b20cd4af23d38653d3f258 .

Arecsu:
Hey glad you ACK this problem as well. Working pretty good and your code is far better than mine. Thank you!

PR #100, opened 2026-06-04
fix: bind webserver dual-stack so IPv4 and IPv6 clients both reach it

urwrstkn8mare:
## Problem

Avahi advertises the host on both IPv4 (A record) and IPv6 (AAAA record) by default, but the web server bound to `0.0.0.0` — IPv4 only. When Moonlight discovers the host via mDNS it gets both addresses and polls `/serverinfo` on each: the IPv4 poll succeeds (host shows online / padlock), while the IPv6 poll is refused (host shows offline / warning triangle). The client rotates between the two, so the host icon flip-flops between online and offline and pairing fails. This makes an auto-discovered Moonshine instance hard to connect to.

This reproduces on any machine with an IPv6 address (e.g. a router-assigned ULA) and `use-ipv6=yes` in Avahi (the default).

## Fix

- Bind the HTTP and HTTPS listeners via `socket2` with `IPV6_V6ONLY` disabled, so a single `::` socket serves both IPv4 and IPv6.
- The default `address` stays `0.0.0.0` (IPv4-only); setting it to `::` opts into dual-stack. Either way the webserver, RTSP, and streams follow the configured address consistently, so discovery and reachability can't disagree anymore.
- Bracket IPv6 hosts in the RTSP `sessionUrl0` (`rtsp://[::1]:48010`) so launch/resume URLs are valid over IPv6.
- Normalize IPv4-mapped addresses (`::ffff:a.b.c.d` → `a.b.c.d`) so MAC lookups and session URLs stay clean for IPv4 clients on the dual-stack socket.

## Testing

- `ss` shows the listeners on `*:47989` / `*:47984`, and `/serverinfo` returns `200` over both the IPv4 and the IPv6 (ULA) address. Before, the IPv6 request was refused. 

## Concerns

With the IPv4-only default, IPv6 reachability is opt-in via `address = "::"`. Pairs well with #111, which advertises only the addresses matching `config.address` — so the default no longer advertises an unreachable AAAA record, and `::` gives working dual-stack end to end (granted no network issues).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2425,6 +2425,7 @@ dependencies = [
  "shellexpand",
  "shlex",
  "smithay",
+ "socket2 0.5.10",
  "steamlocate",
  "strum_macros",
  "tempfile",
--- a/moonshine-core/Cargo.toml
+++ b/moonshine-core/Cargo.toml
@@ -52,6 +52,7 @@ hyper = { version = "1.8.1", features = ["server", "http1"] }
 hyper-util = { version = "0.1.20", features = ["tokio"] }
 http-body-util = "0.1.3"
 network-interface = "2.0.5"
+socket2 = "0.5.10"
 url = "2.5.8"
 notify-rust = "4.12.0"
 
--- a/moonshine-core/src/config.rs
+++ b/moonshine-core/src/config.rs
@@ -94,9 +94,9 @@ impl Default for Config {
 	fn default() -> Self {
 		Self {
 			name: "Moonshine".to_string(),
-			// Bind dual-stack by default so clients can reach us over IPv4 or IPv6.
-			// The webserver disables IPV6_V6ONLY, so this single address covers both.
-			address: "::".to_string(),
+			// IPv4-only by default; set to `::` to bind dual-stack (the webserver
+			// disables IPV6_V6ONLY, so that single address covers both).
+			address: "0.0.0.0".to_string(),
 			webserver: Default::default(),
 			stream: Default::default(),
 			applications: vec![ApplicationConfig {
--- a/moonshine-core/src/config.rs
+++ b/moonshine-core/src/config.rs
@@ -17,6 +17,31 @@ fn default_launch_timeout() -> u64 {
 	2
 }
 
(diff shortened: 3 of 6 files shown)


hgaiser:
I made some changes to this PR. I updated it to the latest main commit and I removed the option of having IPv6 only. To me it makes sense to have either IPv4 or dual-stack.

Could you have a look at the PR to check if everything is still okay?

urwrstkn8mare:
The problem was that the server had to be able to accept both ipv4 and ipv6 API requests from moonshine as avahi-daemon advertises both addresses no matter what (the only way to fix this is if we embedded our own mdns provider via https://github.com/keepsimple1/mdns-sd). However, at least on my network, ipv6 was unstable so I needed to have an option to be able to receive API requests on both but have Moonshine *explicitly* steer the RTSP stream itself to ipv4.

So sorry if this means having to revert that commit but I think leaving it to a sensible default so not many people have to think about it would be better? What do you think? Maybe I should make the config more clear with more comments?

hgaiser:
> The problem was that the server had to be able to accept both ipv4 and ipv6 API requests from moonshine as avahi-daemon advertises both addresses no matter what (the only way to fix this is if we embedded our own mdns provider via https://github.com/keepsimple1/mdns-sd). However, at least on my network, ipv6 was unstable so I needed to have an option to be able to receive API requests on both but have Moonshine _explicitly_ steer the RTSP stream itself to ipv4.
> 
> So sorry if this means having to revert that commit but I think leaving it to a sensible default so not many people have to think about it would be better? What do you think? Maybe I should make the config more clear with more comments?

I've been wanting to replace avahi with something like mdns-sd anyway to reduce the amount of runtime dependencies. Sounds like a good reason to finally do that :)

urwrstkn8mare:
Ooo ok I can work on doing that if you'd like. The only reason I didn't go straight to that is that there could be some issues with multiple mdns providers on one host.

urwrstkn8mare:
@hgaiser btw this PR I reckon is still worth merging incase someone wanted to try out ipv6 capabilities.

hgaiser:
Absolutely, thanks again! Also congratulations on PR #100 :)

PR #103, opened 2026-06-04
fix: allow reconnecting/resuming a running stream

urwrstkn8mare:
Reconnecting to an already-running session failed: the session either rejected the resume outright, or — after an abrupt disconnect — refused the new control connection with a "network issues" error in Moonlight. This fixes both.

> **Depends on hgaiser/tokio-enet#2** (adds `HostConfig::replace_peer_when_full`). Draft until that lands; the `tokio-enet` git rev here points at that PR's branch and should be repointed to upstream once it's merged.

## Problem

The session is a one-way state machine (`Initialized → Launched → Active`) with no path for a client to rejoin once `Active`:

1. On reconnect Moonlight re-runs the full RTSP handshake, but `set_stream_context` (ANNOUNCE) and `start_session` (PLAY) both reject the `Active` state, so the resume gets a 500 and fails.
2. Even past that, the ENet control host uses `peer_count: 1`. After an abrupt disconnect the previous peer stays `Connected` until ENet's reliable-ping timeout (~30s) and keeps the only slot, so a reconnecting client can't get a control channel — ENet drops the incoming CONNECT before it surfaces as an event, and rapid reconnects exhaust the slot ("network issues").

## Fix

- Treat a re-ANNOUNCE / re-PLAY on an `Active` session as a resume no-op. The video, audio, and control streams keep running across the disconnect — the UDP streams re-learn the client's address from its PINGs and keys are refreshed via `/resume` — so there is nothing to rebuild and the game is never interrupted.
- Set `replace_peer_when_full: true` on the control host (new tokio-enet option) so a reconnecting client resets the stale peer and takes over the single slot immediately, instead of being refused.

## Testing

- Disconnect from a running stream and reconnect/resume: reattaches without interrupting the game.
- Abrupt disconnects (force-quit / network drop) and rapid repeated reconnects: no longer wedge or error.
- Setup: Arch, NVIDIA, streaming to an iPad.

### DIFF
--- a/moonshine-core/src/session/manager.rs
+++ b/moonshine-core/src/session/manager.rs
@@ -357,8 +357,13 @@ impl SessionManager {
 				Some(SessionState::Active(active)) => {
 					// Resume (reconnect): the streams are already running, so PLAY is a
 					// no-op — the client picks up the existing streams once it PINGs.
+					// The reconnecting client is a fresh Moonlight session that expects
+					// frame numbers to start at 1, so reset the video frame counters and
+					// force an IDR; otherwise it sees the running counter as a huge frame
+					// gap and reports a poor connection.
+					active.reset_video_stream();
 					guard.session = Some(SessionState::Active(active));
-					tracing::info!("Resuming active session: streams already running, treating PLAY as no-op.");
+					tracing::info!("Resuming active session: resetting video frame counter and treating PLAY as no-op.");
 					return Ok(());
 				},
 				None => {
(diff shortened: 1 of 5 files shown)


hgaiser:
Thanks for all the work! I changed the PR to set `peer_count=128`, which matches what [sunshine](https://github.com/LizardByte/Sunshine/blob/05cadf3867100417628e75a428c71c764f0d8d75/src/network.cpp#L177) does, instead of changing the tokio-enet interface.

PR #104, opened 2026-06-05
fix: stop session when launched app exits

urwrstkn8mare:
Quitting Steam from inside a stream left Moonlight connected to a black screen. The application process had exited, but Moonshine kept the compositor and streams alive until the client explicitly sent `/cancel` or a stream timeout fired.

## Problem

Applications are launched as transient user systemd units (`moonshine-session.service`). Launch already checked that the unit survived startup, but after launch Moonshine dropped the D-Bus subscription and only kept an `Application` value so the unit could be stopped when the session was dropped.

That made the lifecycle one-way: session shutdown stopped the app, but the app exiting did not stop the session. Closing Steam therefore removed the visible content while the compositor/video/audio/control streams continued running, which appears in Moonlight as a black screen that never ends automatically.

## Fix

- Keep a small monitor task on the launched application unit after startup succeeds.
- Watch for `ActiveState=inactive`, `ActiveState=failed`, or `UnitRemoved` and trigger session shutdown with a new `ApplicationStopped` reason.
- Abort the monitor when the session is intentionally stopped, so `/cancel` and normal teardown continue to own app cleanup.

This does not require Moonshine itself to run under systemd; it only watches the transient user unit Moonshine already creates for the launched app.

## Testing

- `cargo build -p moonshine-core`
- `cargo test -p moonshine-core`
- `cargo build --release -p moonshine-core`

Runtime case to verify: launch Steam via Moonlight, quit Steam fully, and the Moonlight session should end instead of staying on a black screen.

### DIFF
--- a/moonshine-core/src/session/application.rs
+++ b/moonshine-core/src/session/application.rs
@@ -1,13 +1,16 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
+use async_shutdown::ShutdownManager;
 use futures_util::StreamExt;
 use std::collections::HashMap;
+use tokio::task::JoinHandle;
 use zbus::proxy::SignalStream;
 use zbus::{Connection, MatchRule, MessageStream, Proxy};
 use zvariant::OwnedObjectPath;
 
 use crate::config::ApplicationConfig;
+use crate::session::manager::SessionShutdownReason;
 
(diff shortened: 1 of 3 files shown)


hgaiser:
Thanks!

PR #109, opened 2026-06-10
Feature: Pass client screen infos as env args to host

BigDiaB:
I don't know how widely applicable/helpful this is to other users, but this is a must have for me.
Some of my steam games (e.g. Forza Horizon 6) realy do not play well with the smithay compositor that is builtin to moonshine.

Forza creates multiple windows, one being black and the other displaying the actual game, but the focus mechanism in moonshine pics the black window.
This may be only an issue on my machine, as a friend's pc with same brand components does not do this.

In my case the usage of Gamescope in Steam's launch args fixes this. But I am frequently switching between different resolution devices, so hardcoding display infos is not a viable option; for me at least.

Hence, this PR that communicates just that using environment variables.

### DIFF
--- a/moonshine-core/src/session/application.rs
+++ b/moonshine-core/src/session/application.rs
@@ -107,6 +107,8 @@ pub(crate) struct ApplicationContext {
 	pub wayland_display: String,
 	/// Effective HDR mode — `true` only when the compositor confirmed an HDR-capable DMA-BUF format is in use.
 	pub hdr: bool,
+	/// Environment variables to pass on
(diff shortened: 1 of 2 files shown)


BigDiaB:
Forgot to mention that Gamescope needs to have internal and actual resolution passed as args to work properly. This would all have been unnecessary for me if Gamescope would just use the fullscreen resolution for both by default.

hgaiser:
Apologies for the delay, vacations and all :)

Thanks!

PR #111, opened 2026-06-11
feat: replace avahi/zeroconf with embedded mdns-sd responder

urwrstkn8mare:
## Problem

Discovery currently goes through the `zeroconf` crate, which is a binding to avahi — so avahi is both a build dependency and a hard runtime dependency. In the review of #100 the preference was to drop avahi entirely rather than keep working around it.

The main reason for me working on this is that there is also a discovery-side half of the flip-flop bug from #100: avahi advertises both A and AAAA records by default, while the webserver only listened on IPv4, so Moonlight alternates between a working IPv4 address and a dead IPv6 one. (As detailed in #100) this was super annoying in Moonlight and made it hard to connect sometimes.

## Fix

- Replace `zeroconf`/avahi with the pure-Rust `mdns-sd` responder (plus `gethostname`). No external daemon needed; avahi is removed from the README and CI package lists (`clang` stays — other crates need it).
- Advertised addresses follow `config.address`:
  - `::` → all interface addresses (IPv4 + IPv6)
  - `0.0.0.0` → IPv4 only
  - a specific IP → just that IP
  
  Since the default (`0.0.0.0`) now advertises IPv4 only, this fixes the online/offline flip-flop that #100 also fixed(ish). It's complementary to that PR (which fixes the webserver-binding half) — they merge cleanly in either order, and this PR doesn't touch config defaults or the webserver.  
- Coexistence with a running avahi: the SRV target is a unique hostname (`<machine>-moonshine.local.`) instead of the machine hostname avahi already owns, so avahi never sees a conflicting unique record. Sharing UDP 5353 is fine (both use `SO_REUSEPORT`), and a shared PTR record is explicitly allowed by RFC 6762. Verified Moonlight resolves whatever SRV target we publish.

## Known limitations

- If avahi is configured with `disallow-other-stacks=yes`, the socket can't be bound: Moonshine logs an error and runs with discovery off (manual add in Moonlight still works). That option is off by default.
- `mdns-sd` sends multicast-only responses (no legacy unicast/QU support). Moonlight clients are fine with this, but it means `dig @224.0.0.251` style queries time out — test with a multicast listener instead.

## Testing

Verified locally (CachyOS, avahi running alongside):

- `0.0.0.0` advertises IPv4 only; `::` advertises dual-stack.
- No avahi conflicts or hostname renames in the journal while both share port 5353.
- Goodbye packets on SIGINT — the service vanishes from `avahi-browse` immediately.
- The responder answers raw multicast PTR queries for `_nvstream._tcp.local.`.

### DIFF
--- a/moonshine-core/Cargo.toml
+++ b/moonshine-core/Cargo.toml
@@ -56,8 +56,8 @@ url = "2.5.8"
 notify-rust = "4.12.0"
 
 # Discovery
-gethostname = "1.0"
-mdns-sd = "0.20"
+gethostname = "1.1.0"
+mdns-sd = "0.20.0"
 
 # Session (compositor)
 ash = { git = "https://github.com/ash-rs/ash", rev = "55dd56906bbb5760e9e9e6c56f45be67f67e0649" }
--- a/moonshine-core/Cargo.toml
+++ b/moonshine-core/Cargo.toml
@@ -56,7 +56,8 @@ url = "2.5.8"
 notify-rust = "4.12.0"
 
 # Discovery
-zeroconf = "0.17.0"
+gethostname = "1.0"
+mdns-sd = "0.20"
 
 # Session (compositor)
 ash = { git = "https://github.com/ash-rs/ash", rev = "55dd56906bbb5760e9e9e6c56f45be67f67e0649" }
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -6,7 +6,6 @@ runs:
     - name: Install dependencies
       run: |
         sudo pacman -Syyuu --noconfirm --needed \
-          avahi \
           clang \
           cmake \
           gcc-libs \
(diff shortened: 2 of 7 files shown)


moonshine-core/Cargo.toml
```diff
@@ -56,7 +56,8 @@ url = "2.5.8"
 notify-rust = "4.12.0"
 
 # Discovery
-zeroconf = "0.17.0"
+gethostname = "1.0"
```
hgaiser: I prefer to list full version numbers.

hgaiser (commented):
One minor comment, but looks good to me!

hgaiser:
I see you changed the version numbers; I'm ready to merge when you are.

urwrstkn8mare:
@hgaiser it turns out while the discovery seemed to be working moonlight itself still exhbitted an issue where instead of flip flopping it more just flipped (i.e. found moonlight showed padlock then went offline forever-ish). turns out the fix for this was just returning the local ip in /serverinfo. I wonder if this change alone would have fixed my issue without mdns-sd but since you wanted to go over to it anyways we might as well. 

PR ready to merge when you are.

PR #112, opened 2026-06-11
fix: bench hanging forever when no frames arrive before the duration deadline

urwrstkn8mare:
If the app never delivers a frame, the duration timeout's `duration_remaining.as_secs() > 0` guard stops passing once the remaining time drifts below a second, and the bench spins on a zero-length sleep forever instead of exiting. Checking the clock against the deadline directly makes it always stop at `--duration`.

This was mainly just as a result of a hang I had randomly that confused me while using moonshine-bench (which is very cool btw).


hgaiser:
Thanks! Awaiting CI and then I'll merge.

PR #115, opened 2026-06-15
fix: terminate orphaned XWayland processes on session shutdown

### LINKED ISSUE #89: XWayland processes lingering after session terminates
At the end of a streaming session, all processes related to the session exit except one.
It looks like this:
Xwayland :0 -verbose -rootless -terminate -wm 78 -displayfd 85 -listenfd 82 -listenfd 83

The ":0" increments, thus after n sessions we have "Xwayland :1", "Xwayland :2", ... "Xwayland :n" processes.
Each of those processes consumes some RAM and VRAM (see nvtop listing in image below) 

Image

me-teorick:
Fixes #89, although as discussed it is merely a workaround while waiting for Smithay to fix the issue upstream (see smithay/smithay#1210).

### DIFF
--- a/moonshine-core/src/session/compositor/state.rs
+++ b/moonshine-core/src/session/compositor/state.rs
@@ -8,6 +8,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc;
 use std::sync::Arc;
 
+use libc;
+
 use smithay::reexports::wayland_server::Resource;
 
 use smithay::backend::allocator::dmabuf::{AsDmabuf, Dmabuf};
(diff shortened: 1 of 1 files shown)


hgaiser:
I appreciate the effort, but since there's no direct issue caused by the orphaned Xwayland processes I want to see if we can fix this in Smithay first. 

When https://github.com/hgaiser/moonshine/pull/108 gets merged we'll be using an up-to-date Smithay anyway, so any fix on top of Smithays `main` will be pulled in easily.

Do you want to submit a PR to Smithay instead? I wouldn't mind a Moonshine PR to patch the Smithay dependency if you have a working version, even if it's not merged yet.

hgaiser:
I'm going to close this PR in favor of some fix in Smithay. It's pretty harmless to keep those processes running in the background and I don't want to clutter the code with cleaning up an issue that should be resolved in Smithay.

Regardless, thanks for the input!

PR #117, opened 2026-06-28
fix: use platform-agnostic c_char type instead of i8

neobrain:
On some platforms (notably ARM), char is unsigned rather than signed, so the build would fail due to use of i8. The FFI type alias works everywhere.


hgaiser:
Thanks!

PR #123, opened 2026-07-21
Signal and convert SDR as limited-range BT.709

plasticchris:
SDR was encoded as full-range BT.709 in both the VUI color description and the color converter, mismatching the limited (studio) range clients decode as and washing out SDR levels. Use limited range for SDR consistently.


hgaiser:
Thanks, nice catch! I prefer to fix this in [pixelforge](https://github.com/hgaiser/pixelforge/pull/24) though, so I will close this in favor of https://github.com/hgaiser/moonshine/pull/126 .

PR #125, opened 2026-07-21
feat: add Ubuntu .deb packaging, AppImage, Flatpak and GitHub Actions release workflow

mariolameiras:
Add packaging scripts for building Moonshine on Ubuntu/Debian:

.deb package (packaging/deb/build.sh):
- Full systemd system service install (moonshine@.service)
- Wrapper script for default config path (~/.config/moonshine/config.toml)
- udev rules, kernel module config, Vulkan layer
- DEBIAN control, postinst, prerm, postrm, copyright, changelog

AppImage (packaging/appimage/build.sh):
- Self-contained binary via linuxdeploy
- Bundles shared libs (libopus, libevdev, libxkbcommon, libffi)
- System integration helper (install-system.sh)

GitHub Actions (release-ubuntu.yaml):
- Builds .deb and AppImage on ubuntu-24.04 runner
- Triggers on release published, attaches artifacts

New dist/ assets: desktop entry, appstream metainfo, man page, launcher

Tested on Ubuntu 26.04

### DIFF
--- a/.github/workflows/release-ubuntu.yaml
+++ b/.github/workflows/release-ubuntu.yaml
@@ -1,0 +1,58 @@
+name: Release Ubuntu
+
+on:
+  release:
+    types: [published]
+
+env:
+  CARGO_TERM_COLOR: always
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: Build .deb + AppImage
+    runs-on: ubuntu-24.04
+    if: startsWith(github.ref, 'refs/tags/')
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y \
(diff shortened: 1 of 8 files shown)


hgaiser:
Thanks, but I opted to use nfpm in https://github.com/hgaiser/moonshine/pull/130 so that the maintenance will hopefully be low 🤞 


Can you have a look at https://github.com/hgaiser/moonshine/releases to see if the DEB package works for you?

mariolameiras:
Hi @hgaiser yes that makes sense.

The .deb works fine but one thing that I was adding in this PR was a default config.toml file in the user home to make it easier for users to get started.

You could copy that logic over.

Cheers and thank you for your great work!

hgaiser:
How do we choose what user to store the config file under though? The user that installs moonshine is not necessarily the user that runs it. I think it makes more sense to install without a config file and have it be generated at first launch.

PR #128, opened 2026-07-22
Add Fedora RPM packaging

akvartz:
Adds RPM packaging for Fedora under `dist/fedora/`, following the existing `dist/` convention:

- `moonshine.spec` — builds the workspace offline; produces a `moonshine` package (server, `start-moonshine.sh`, systemd unit, udev rules, modules-load config, and the WSI Vulkan layer with its manifest path rewritten for `%{_libdir}`) and a `moonshine-tools` subpackage with `moonshine-bench`.
- `make-source-tarballs.sh` — generates the pristine source tarball plus a `cargo vendor` tarball, needed because the pinned git dependencies (smithay, pixelforge, ash, inputtino) can't be fetched during an RPM build. Includes a fixup for `inputtino-sys`, whose build.rs expects the full inputtino repo two levels above the crate — the script grafts the C sources into the vendored crate.
- `README.md` — build and install instructions.

Verified end-to-end with `rpmbuild -ba` on Fedora 44 (x86_64): packages build, the binary runs, and the Vulkan layer manifest resolves correctly. Intended for COPR/local use rather than official Fedora repos, since Fedora's guidelines don't allow git-pinned dependencies.

### DIFF
--- a/dist/fedora/README.md
+++ b/dist/fedora/README.md
@@ -1,0 +1,61 @@
+# Fedora packaging
+
+RPM packaging for Moonshine. Produces:
+
+- `moonshine` — the streaming server, start script, systemd unit
+  (`moonshine@.service`), udev rules, modules-load config, and the
+  `VkLayer_moonshine_wsi` Vulkan implicit layer.
+- `moonshine-tools` — the `moonshine-bench` encoding-pipeline benchmark.
+
+## Building locally
+
+Install the build dependencies:
+
+```sh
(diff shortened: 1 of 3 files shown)


hgaiser:
Thanks, but I opted to use nfpm in https://github.com/hgaiser/moonshine/pull/130 so that the maintenance will hopefully be low 🤞 


Can you have a look at https://github.com/hgaiser/moonshine/releases to see if the RPM package works for you?

PR #129, opened 2026-07-22
build: add nix flake with package and NixOS module

scottjab:
This adds a nix flake so that NixOS users can build and run Moonshine straight from this repository, plus a `services.moonshine` NixOS module that sets up everything the installation section of the README describes: the systemd service, user lingering, the `uinput`/`uhid` kernel modules, udev rules, and (optionally) the firewall ports. There is also a dev shell with the full build environment, including the `clippy` and `rustfmt` setup CI uses.

Everything lives in `flake.nix`, `flake.lock`, and the `nix/` directory. The only touch on existing files is a short NixOS subsection in the README linking to [nix/README.md](https://github.com/hgaiser/moonshine/blob/main/nix/README.md), which documents usage in the same style as the rest of the README. Nothing about the existing build or CI changes.

## Why in this repo instead of a fork

Flakes are consumed by repository URL, so having it here means `nix build github:hgaiser/moonshine` and the module import just work for everyone, without anyone maintaining a separate packaging repo that drifts out of date.

## Maintenance cost if you don't use nix

This was written so it should be none in practice:

- The nix build derives all dependency vendoring from `Cargo.lock`. Bumping, adding, or removing crates (including the git dependencies) needs no changes on the nix side.
- The one exception is a pinned hash for `ash`, because its locked rev sits on an unmerged branch that ref-based fetchers cannot reach. If the ash pin ever changes, the nix build fails loudly and prints the correct new hash (and once ash points at a normal release, the entry just gets deleted). The comment in `nix/package.nix` explains this.
- New system library dependencies (new `-sys` crates) would need a one-line addition, which nix users will notice and can PR.
- `flake.lock` pins nixpkgs and only needs occasional bumping, which again nix users can PR.

Verified with a full `nix build` on x86_64-linux, and the module is running Moonshine on a headless NixOS host streaming Steam.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 __pycache__/
+/result
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ To run Moonshine for your user:
    sudo systemctl enable --now moonshine@$USER
    ```
 
+### NixOS
+
+This repository is also a nix flake, providing a package and a NixOS module that sets up the service for you.
+See [nix/README.md](nix/README.md) for instructions.
+
 ### Source
 
 The following dependencies are required to build and run:
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1784497964,
+        "narHash": "sha256-vlHUuqAcbcH2RKmHbPiuQzbv1pnzzavXnI62RD0bqCU=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "241313f4e8e508cb9b13278c2b0fa25b9ca27163",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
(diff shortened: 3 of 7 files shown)


scottjab:
Thanks again for merging my other PR. 

This is how I am running Moonshine personally, I can totally maintain a wrapper repo if you don't want the nix package stuff in the main repo.

ruffsl:
> There is also a dev shell with the full build environment, including the clippy and rustfmt setup CI uses.

Ooh, awesome! A upstream maintained dev shell will come in real hand when I have to time to directly contribute.

BTW, @scottjab , perhaps you may also like to review the downstream packing/caching Moonshine for nixpks?

- https://github.com/NixOS/nixpkgs/pull/544393
- https://github.com/NixOS/nixpkgs/pull/544596

hgaiser:
Let's see how much maintenance this will turn out to be 🙂

I understand the convenience of having it in this repo, so I'll merge it, though I have no experience with Nix or its packages. Thanks for submitting this 👍

scottjab:
Feel free to assign nix PRs to me for now.

PR #137, opened 2026-07-26
fix: validate DMA-BUF import cache entries with kcmp(2)

oliverlj:
### Problem

`DmaBufImporter` caches `VkImage`/`VkDeviceMemory` per DMA-BUF fd and reuses an entry when
`ImportParams` (width, height, format, modifier, plane offsets/strides) match.

Kernel fd numbers are recycled: when a `wl_buffer` is released and its fd closed, the same
number is handed back for an unrelated buffer. `ImportParams` cannot catch this, because
swapchain images are identical on every field it compares. The cache therefore returns an image
bound to a *previous* buffer's memory, and the encoder re-encodes a stale frame — intermittently,
for as long as the session runs.

### Symptom

A permanent shimmer over the whole picture during direct scanout, as current and stale frames
alternate. It is game-dependent, which makes it look random: whether an fd gets recycled onto a
live cache entry depends entirely on how fast a game turns over its `wl_buffer`s.

Reproduced on an NVIDIA host streaming 4K60 HDR (HEVC Main 10), direct scanout active:

| Game | Result |
|------|--------|
| FINAL FANTASY VII REMAKE INTERGRADE | shimmers permanently |
| FINAL FANTASY VII REBIRTH | clean |

Both present through the same path with **identical** buffer parameters —
`fourcc=XB30`, `modifier=0x300000000e08014`, 1 plane, 3840x2160 — so the difference is not in
what is imported, only in when fds get reused.

### Fix

Keep the fd as the cache key, and give each entry its own `dup` of that fd, so an incoming fd can be
checked against it with `kcmp(2)` before the entry is reused. All fds for one dma-buf are dups of the
exporter's single `struct file`, so "same open file description" is exactly "same buffer" — as
opposed to "same slot in the process file table", which is all the key itself guarantees.

Where the kernel refuses `kcmp` (no `CONFIG_CHECKPOINT_RESTORE`, container seccomp), fall back to
comparing the dma-buf's inode; the entry's own dup pins the buffer, so that inode cannot be recycled
while the entry lives. Reuse is refused when neither check can answer, rather than risking a stale
image.

Cost is one syscall per import call. The cache keeps working normally — measured hit rate 99 %.

### What was ruled out

Worth recording, since several of these look plausible:

- **Missing acquire fence** (encoder reading a partially rendered frame). Measured with
  `DMA_BUF_IOCTL_EXPORT_SYNC_FILE` on the scanout buffer over 4800 frames: `absent=0`,
  `pending=0`, `signalled=4800`. Client rendering is always complete when the encoder reads.
  A fence would have changed nothing.
- **Buffer format / modifier mismatch.** Measured identical between the affected and unaffected
  game (see table above).
- **Present mode.** Forcing FIFO through the frame-limiter path made no difference.
- **Client-side (moonlight-qt) rendering.** Unaffected by the server-side toggle that makes the
  symptom appear and disappear.

### An approach that looks right but is not

Keying the cache on the compositor's stable per-`wl_buffer` index (`scanout_buffer_map`) also
removes the shimmer — but only because it disables the cache: NVIDIA wraps a new `wl_buffer`
every frame, so every frame gets a fresh index and therefore a full Vulkan import — the thousands of
cache hits measured above would all have become full imports. The identity check keeps the cache
working.

### Testing

Built against `v0.13.5`. Shimmer gone on the affected game, unaffected game still clean, HDR
colours correct, automatic SDR↔HDR switching unaffected.

### DIFF
--- a/moonshine-core/src/healthcheck.rs
+++ b/moonshine-core/src/healthcheck.rs
@@ -145,7 +145,6 @@ fn iter_checks(report: &HealthReport, mut f: impl FnMut(CheckOutcome, &str, &str
 	}
 }
 
-
 /// Run the health check and log the results to the console.
 pub fn print_health_report(report: &HealthReport) {
 	let tty = std::io::stdout().is_terminal();
@@ -276,6 +275,7 @@ pub fn run_healthcheck(config: Option<&Config>) -> HealthReport {
 	check_xwayland(&mut report);
 	check_xdg_runtime(&mut report);
 	check_inhibit(&mut report);
+	check_kcmp(&mut report);
 
 	// --- Ports (requires config) ---
 
@@ -817,6 +817,62 @@ fn check_inhibit(report: &mut HealthReport) {
 		"Can acquire logind sleep inhibit (user in moonshine group, polkit rules ok)".into(),
 		start.elapsed().as_millis() as u64,
 	);
+}
+
+fn check_kcmp(report: &mut HealthReport) {
+	let start = Instant::now();
+
+	/// `KCMP_FILE` — not exposed by `libc` or glibc.
+	const KCMP_FILE: libc::c_int = 0;
+
+	// SAFETY: uses /dev/null fds; syscall with scalar arguments.
+	let result = unsafe {
+		let a = std::fs::File::open("/dev/null");
+		let b = std::fs::File::open("/dev/null");
+		match (a, b) {
+			(Ok(a), Ok(b)) => libc::syscall(
+				libc::SYS_kcmp,
+				libc::getpid(),
+				libc::getpid(),
+				KCMP_FILE,
+				a.as_raw_fd() as libc::c_ulong,
+				b.as_raw_fd() as libc::c_ulong,
+			),
+			(Err(e), _) | (_, Err(e)) => {
+				report.add_failed(
+					"kcmp(2)",
+					format!("  Cannot open /dev/null for kcmp(2) probe: {e}"),
+					start.elapsed().as_millis() as u64,
+				);
+				return;
+			},
+		}
+	};
+
+	if result < 0 {
+		let error = std::io::Error::last_os_error();
+		let hint = match error.raw_os_error() {
+			Some(libc::ENOSYS) => {
+				"  Kernel was compiled without CONFIG_CHECKPOINT_RESTORE.\n  \
+				kcmp(2) is required for DMA-BUF import cache validation.\n  \
+				Rebuild your kernel with CONFIG_CHECKPOINT_RESTORE=y."
+			},
+			Some(libc::EPERM) => {
+				"  kcmp(2) is blocked by a seccomp filter (e.g. Flatpak/container).\n  \
+				kcmp(2) is required for DMA-BUF import cache validation.\n  \
+				Run outside the sandbox or adjust the seccomp profile."
+			},
+			_ => "  kcmp(2) is required for DMA-BUF import cache validation.",
+		};
+		report.add_failed(
+			"kcmp(2)",
+			format!("  kcmp(2) not available: {error}\n{hint}"),
+			start.elapsed().as_millis() as u64,
+		);
+		return;
+	}
+
+	report.add_passed("kcmp(2)", "Available".into(), start.elapsed().as_millis() as u64);
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


hgaiser (commented):
Cool, thanks for the thorough analysis. I don't like the existing method of keying DMA-BUF so I am definitely open to changing it. I was reading into this and found [`kcmp`](https://man7.org/linux/man-pages/man2/kcmp.2.html), which sounds like it's exactly what we need. Could you change the logic to use `kcmp` for file descriptor comparison and check if you identify recycled frames in your games again?

I would do it but I'm not sure I have a game that exhibits this problem. Or at the very least I never ran into this yet.

oliverlj:
Hi, I used claude to generate the pr. I don't know rust. ff7 remake works perfect now with this patch.

Here is the report : 

Done — the branch now validates entries with kcmp(2).

  One consequence worth flagging, because it changes the shape of the patch: kcmp(…, KCMP_FILE, …) compares two live descriptors, and the importer had none to compare against — import_internal dups the incoming
  fd and hands the dup to vkAllocateMemory, which takes ownership of it. So each cache entry now keeps its own dup for its lifetime, and the check is kcmp(getpid(), getpid(), KCMP_FILE, entry.fd, incoming_fd) ==
  0. One extra fd per live entry (a handful at 4K60), and one syscall per import call — the same cost as the fstat it replaces.

  It is also a better question to ask than the inode one: all fds for a dma-buf are dups of the exporter's single struct file — dma_buf_fd() installs dmabuf->file, and SCM_RIGHTS passes that same file across the
  socket — so "same open file description" is exactly "same buffer", with no reasoning needed about when inode numbers get reused.

  Fallback. kcmp needs CONFIG_CHECKPOINT_RESTORE, and Docker's default seccomp profile blocks it without CAP_SYS_PTRACE. Rather than silently losing the cache there (0.5–1.5 ms per frame back on the hot path), I
  kept the inode comparison as a fallback, latched on the first ENOSYS/EPERM. Holding our own dup pins the buffer, so its inode cannot be recycled while the entry lives, which makes the fallback sound rather
  than best-effort. Happy to drop it and just refuse reuse when kcmp is unavailable if you would rather keep the patch minimal — it is ~20 lines.

  Measurements. I ran an instrumented build that evaluates both verdicts — kcmp and the inode comparison — on every candidate hit, and counts the cases where they disagree, so the two methods can be checked
  against each other over real sessions. I also split the retirement counter into recycled (parameters matched, so the fd number genuinely moved to another buffer — the case the current code serves stale) and
  reconfigured (parameters differed, which ImportParams already catches today).

  FINAL FANTASY VII REMAKE INTERGRADE, 4K60 HDR, direct scanout, NVIDIA:

  hits=64790  misses=10  recycled=2  reconfigured=0  kcmp_only=0  inode_only=0  hit_rate_pct=99

  Three things came out of it:

  - Recycled fds are still identified, and the trigger turns out to be deterministic rather than random. Both events land in the same 17 ms, right as the game enters HDR:

  16:13:06.680  Input format changed from RGBA16F to ABGR2101010, recreating color converter
  16:13:09.183  swapchain_feedback: HDR10_ST2084 detected, activating HDR
  16:13:09.196  recycled fd, identical params: fd=138
  16:13:09.213  recycled fd, identical params: fd=141

  - The HDR transition releases the old buffers, and their fd numbers come straight back onto cache entries that are still live. fd=141 was also the number caught in an earlier session — the allocator reuses the
  same low slots.
  - reconfigured=0. The ImportParams guard caught nothing at all, in any session. The format change itself does not retire anything, because the new buffers arrive on fresh fds. Every retirement came from the
  identity check.
  - kcmp_only=0 inode_only=0. The two methods returned the same verdict on every single candidate hit — ~65k here, and ~668k over a separate 3-hour session that produced one recycled fd and likewise zero
  disagreements. So this is the same defect being caught, just identified properly.

  Cache efficiency is unaffected at 99 %, and the picture stayed clean for the whole session — including after the two recycled fds, which is the point: before the fix, a single poisoned entry kept being handed
  back and made current and stale frames alternate for the rest of the session.

hgaiser:
Thanks for the PR! I made some changes; I tweaked the docstrings, removed inode checking entirely (kcmp is available on pretty much all Linux kernels) and added a check to the health check. Everything should still work, but if you could verify that, that would be great!

oliverlj:
Hi, thanks for the work.

I have tested successfully.

hgaiser:
Cool, thanks for the feedback.

Also I like your profile picture, brings back memories :p

Issue #113, opened 2026-06-14
[Feat] Explore using UDP Generic Segmentation Offload (GSO)


urwrstkn8mare:
This could reduce the number of syscalls in our hotpath and allow the NIC/kernel to handle the batching of packets instead of our own. Some things to consider, though, are pacing as, at least on Wifi in my early testing, I seemed to get more packet loss as bursts of packets were unable to  be dealt with by my iPad. I believe Sunshine uses this. Should be worth exploring.

hgaiser:
I had a stab at trying GSO to reduce syscalls. I got the following results:

| | Before (per-shard `sendto`) | After (GSO `sendmsg`) | Reduction |
|---|---|---|---|
| **`sendto` calls** | **8,967** | 732 (non-video only) | **92%** |
| **`sendmsg` calls** | 83 (non-video) | **628** (GSO video + non-video) | n/a |
| **Total syscalls** | **9,050** | **1,360** | **85%** |
| **Video syscalls** | ~8,884 `sendto` | ~545 `sendmsg` | **94%** |
| **Syscalls per frame** | ~18.8 | ~1.3 | **93%** |
| **Syscall CPU time** | 32ms | 11ms | **66%** |

So there was a reduction of ~21msec over a period of ~12s. So CPU time was reduced by about ~1.8ms/sec. This was while capturing with `strace`, so there might even be some inflation of values because of that.

I submitted a PR in https://github.com/hgaiser/moonshine/pull/138 in case you want to have a look.

Overall the results are clear, the difference in measurable latency not so much :). Still though, the added code is quite small so I'm inclined to add it anyway.

hgaiser:
Fixed by #138 and #157

urwrstkn8mare:
wow very cool 🎊

PR #138, opened 2026-07-27
feat: use Generic Segmentation Offload in video stream

hgaiser:
PR for https://github.com/hgaiser/moonshine/issues/113


hgaiser:
I'm going to merge this. The gain is quite small but the code added is also quite small 🤷

Issue #46, opened 2026-03-27
SDR content appears washed out when HDR is enabled


jfernandez:
When streaming SDR content (like Steam Big Picture) with HDR enabled in the Moonlight client, colors look noticeably washed out compared to running the same content natively. Disabling HDR on the client fixes the colors, but you obviously lose HDR for games that actually use it.

**Setup:** Host is an RTX 5080 running Bazzite desktop, client is a Legion Go 2 running Bazzite with moonlight-qt 6.1.0

The issue is that moonshine unconditionally converts all frames through sRGB -> BT.2020 PQ when the client negotiates HDR, even when the content is plain SDR. The client's libplacebo then tonemaps the PQ signal back to the display output. This round-trip is lossy. I tried various metadata values (including `maxDisplayLuminance=203` to collapse tonemapping to identity) and fixing the color range mismatch (`full_range: false` vs the client's forced `PL_COLOR_LEVELS_FULL`), but nothing fully eliminated the visual difference. The PQ round-trip through libplacebo doesn't produce an identity result for SDR content.

I was able to fix this by keeping 10-bit Main10 encoding for HDR sessions but skipping the PQ conversion for SDR frames, encoding them as 10-bit BT.709 instead, and letting libplacebo handle SDR-in-HDR display rendering natively (the same path mpv uses). The encoder is recreated on color space transitions because the Vulkan video session params are immutable, and `hdr_mode` is now driven by the actual frame content rather than being sent unconditionally at session start. I'm happy to submit a PR if this approach makes sense.

PR #116, opened 2026-06-16
feat: "pushified" pipelined video encoding with pixelforge

urwrstkn8mare:
companion PR to https://github.com/hgaiser/pixelforge/pull/21 (which means it depends on it which is why it is a draft for now)

PR #146, opened 2026-07-29
fix(video): honour the client's requested color range

### LINKED ISSUE #145: [Bug] Raised blacks for whole stream compared to Sunshine
Blacks are visibly raised compared to the same stream vs Sunshine on the same client and same TV. Very visible even on just Steam Big Picture startup:

- Native SteamOS on SteamMachine, loading - Steam logo is white, the rest if the picture is pitch black
- Sunshine with Steam Big Picture - same
- Moonshine - Steam logo is white, the rest of the picture is grey-ish

I'm not sure I saw this before on other clients, but certainly looks off streaming to Steam Machine.

Moonshine 0.14.1.

lutyjj:
Fixes #145.

We hardcode `full_range = false` in the color converter and the encoder VUI, and never read `x-nv-video[0].encoderCscMode`. Some clients ask for full range (mode 3) and decode accordingly, so our limited-range data lands with black at 16 instead of 0.

Sunshine reads bit 0 of the same attribute:

```cpp
colorspace.full_range = (config.encoderCscMode & 0x1);
```

Was noticed before in #46 (`full_range: false` vs the client's forced `PL_COLOR_LEVELS_FULL`), and doesn't look like it got fixed in #47.

The remaining bits pick an SDR colorspace. We always encode Rec.709, which is fine as long as the VUI says so, so I left that alone but log a warning on an explicit non-709 request. An absent attribute isn't treated as a request.

Steam logo used to be white on a grey background, now white on proper black.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2846,8 +2846,8 @@ dependencies = [
 
 [[package]]
 name = "pixelforge"
-version = "0.7.2"
-source = "git+https://github.com/hgaiser/pixelforge?rev=9d9934c44077c89ec56f1d2a6667b05ff563fdc4#9d9934c44077c89ec56f1d2a6667b05ff563fdc4"
+version = "0.8.0"
+source = "git+https://github.com/hgaiser/pixelforge?tag=v0.8.0#1767cfc036ca197871f326f4c8e0e71e4b170726"
 dependencies = [
  "ash",
  "futures-channel",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2847,7 +2847,7 @@ dependencies = [
 [[package]]
 name = "pixelforge"
 version = "0.7.2"
-source = "git+https://github.com/hgaiser/pixelforge?tag=v0.7.2#061d5ac05cb8d8257af4842228fc38b14bcb55a8"
+source = "git+https://github.com/hgaiser/pixelforge?rev=9d9934c44077c89ec56f1d2a6667b05ff563fdc4#9d9934c44077c89ec56f1d2a6667b05ff563fdc4"
 dependencies = [
  "ash",
  "futures-channel",
--- a/moonshine-core/Cargo.toml
+++ b/moonshine-core/Cargo.toml
@@ -36,7 +36,7 @@ notify-rust = "4.17.0"
 open = "5.3.5"
 opus = "0.3.1"
 pkcs8 = "0.11.0"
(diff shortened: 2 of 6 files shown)


lutyjj:
Related to pixelforge#24. Limited as the SDR default is right, this just makes us read what the client actually asked for instead of assuming it.

lutyjj:
@hgaiser two things that might belong in pixelforge rather than here:

1. `ColorDescription` has no `with_*` methods, so this uses struct-update syntax at 7 sites. Can add `ColorDescription::with_full_range` there if it'll look better, would just need a pin bump.

2. The other `encoderCscMode` bits (Rec.601, BT.2020 SDR) can't be expressed today: `ColorSpace` has no 601 variant and the shader only carries BT709 and BT2020 matrices. So this just logs a warning if a client asks for one.

lutyjj:
found a regression, putting into WIP for now

something is not right when trying to stream it with HDR AV1. the game starts in SDR regardless, then after the first warning screen it switches to HDR, and whole screen gets green. This seems to be unrelated to this PR though. If you reconnect (or new key frame arrives) the stream recovers.

upd: doesn't seem to be related to PR. Works fine with HEVC.

hgaiser:
> @hgaiser two things that might belong in pixelforge rather than here:
> 
>     1. `ColorDescription` has no `with_*` methods, so this uses struct-update syntax at 7 sites. Can add `ColorDescription::with_full_range` there if it'll look better, would just need a pin bump.
> 
>     2. The other `encoderCscMode` bits (Rec.601, BT.2020 SDR) can't be expressed today: `ColorSpace` has no 601 variant and the shader only carries BT709 and BT2020 matrices. So this just logs a warning if a client asks for one.

Regarding 1: I agree, it makes sense to make a builder pattern for ColorDescription 👍

hgaiser:
@lutyjj do you want to work on the pixelforge PR before merging this?

Also: is_hdr could maybe also be a member function of ColorDescription

lutyjj:
@hgaiser yep, will work on it today/tomorrow :)

lutyjj:
@hgaiser pixelforge side is up: [hgaiser/pixelforge#30](https://github.com/hgaiser/pixelforge/issues/30).

I see though that moonshine is using old pin of pixelforge and would require a little bit more work to unblock this PR :)

hgaiser:
That's fine, it will be updated soon (https://github.com/hgaiser/moonshine/pull/116)

hgaiser:
Just tested this, works good as far as I can tell. Thanks!

PR #147, opened 2026-07-29
fix(xwm): implement X11 fullscreen requests

### LINKED ISSUE #143: [Bug] AC4 Resynced won't start
AC4 Resynced hangs on a grey screen few seconds after the launch.
1. Launches fine
2. Ubisoft Connect screen appears
3. AC4 boot logo appears
4. Grey screen
Process stays alive, occupies 100% of CPU, GPU is idle, so it never gets past the startup.
Game works fine under Sunshine with gamescope on same headless machine.

Client: Steam Machine
Host: headless RTX 5060 Ti, Cachy, Nvidia 610.43.03.

Moonshine 0.14.1.

lutyjj:
Fixes #143.

`XwmHandler::fullscreen_request` / `unfullscreen_request` aren't implemented, so Smithay's no-op defaults run and the request disappears. We advertise `_NET_WM_STATE_FULLSCREEN` in `_NET_SUPPORTED` though, so the client waits for an ack that never comes.

AC4 while stuck:

```
0x7400001   3840x2138+0+0   game window, _NET_WM_STATE = FOCUSED only
  0x6c0009f 3840x2160+0+0   swapchain, still full output size
```

22px short of the output while rendering at full size, then it spins in `d3d12core` at 100% CPU with 0% GPU.

Follows gamescope: `handle_net_wm_state` sets the state and marks focus dirty, geometry gets re-imposed in `determine_and_apply_focus`. So `set_x11_fullscreen` handles state, and `enforce_fullscreen_geometry` holds the focused window at output size during focus evaluation. Correcting there instead of refusing the ConfigureRequest fixes a wrong-sized window whatever caused it, not only when the client asks.

Tested on RTX 5060 Ti / nvidia 610.43.03: AC4 reaches gameplay, window is 3840x2160, `_NET_WM_STATE` has `FULLSCREEN`.

### DIFF
--- a/moonshine-core/src/session/compositor/focus.rs
+++ b/moonshine-core/src/session/compositor/focus.rs
@@ -199,6 +199,13 @@ impl WindowMetadata {
 		self.app_id == 769
 	}
 
+	/// Returns `true` if the window should fill the output. Steam counts even
+	/// without the fullscreen state set.
+	/// Gamescope: `window_is_fullscreen()`
+	pub fn is_fullscreen(&self) -> bool {
+		self.fullscreen || self.is_steam_big_picture()
+	}
+
 	/// Returns `true` if the window has skipTaskbar AND skipPager but is not fullscreen.
 	/// Gamescope: `win_skip_and_not_fullscreen()`
 	pub fn skip_and_not_fullscreen(&self) -> bool {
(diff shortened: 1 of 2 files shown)


hgaiser:
Thanks! Looks good to me. Tested it with a small X11 application which sends this request. Seems to work well 👍

PR #148, opened 2026-07-29
fix(color): do not infer BT.2020+PQ from mastering metadata

### LINKED ISSUE #144: [Bug] Oversaturated colors in AC4 Resynced in SDR mode
Observed on same setup as here: https://github.com/hgaiser/moonshine/issues/143

Steps to reproduce:
- Moonlight - "Enable HDR" on
- SteamOS - HDR off (as HDR broken on SM right now)
- Start AC4 Resynced with HDR off - colors will be oversaturated

Moonshine 0.14.1.

lutyjj:
Fixes #144.

`set_gamescope_hdr_metadata` defaults to BT.2020+PQ when no color space is declared yet. But `vkSetHdrMetadataEXT` carries only mastering metadata, and SDR games call it anyway: DXVK's `DxgiSwapChain::SetHDRMetaData` checks struct type and size, then forwards without looking at the swapchain color space.

The sRGB path also records "this surface is SDR" by removing the entry (`clear_gamescope_current`), which is indistinguishable from "nothing declared yet". So metadata arriving afterwards promotes it:

```
swapchain_feedback vk_colorspace=0 vk_format=37
commit: inserting into current color_space=Srgb
set_hdr_metadata max_cll=2000 max_fall=500
set_gamescope_hdr_metadata color_space=Bt2020Pq
Switching encoder color description to { color_primaries: 9, transfer_characteristics: 16, matrix_coefficients: 9 }
```

The color space and the metadata have separate owners (`swapchain_feedback` and `vkSetHdrMetadataEXT`) and arrive in either order, so track them as two optional fields on a `SwapchainColor` instead of merging both into `ImageDescription`. An image description only exists once a color space was declared, so metadata has no path to imply one. Same pair gamescope keeps in `wlserver_vk_swapchain_feedback`, `vk_colorspace` next to a nullable `hdr_metadata_blob`.

Also added the same validity check gamescope has: metadata with zero light levels or a zero white point gets discarded, we had no check at all.

Tested with HDR off in AC4 Resynced and SteamMachine+Moonlight, stream stays `Srgb` / BT.709.

### DIFF
--- a/moonshine-core/src/session/compositor/color_management.rs
+++ b/moonshine-core/src/session/compositor/color_management.rs
@@ -173,6 +173,63 @@ pub(crate) struct ColorRepresentationSurfaceData {
 // Compositor-level color management state
 // ---------------------------------------------------------------------------
 
+/// Mastering metadata from `vkSetHdrMetadataEXT`, in raw protocol units.
+///
+/// Holds no transfer function or primaries, so it cannot imply a color space.
+/// [`HdrMetadata`] is the u16 wire form sent to the Moonlight client.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct MasteringMetadata {
+	max_cll: u32,
+	max_fall: u32,
+	mastering_luminance: (u32, u32),
+	mastering_primaries: [(u32, u32); 3],
+	white_point: (u32, u32),
+}
+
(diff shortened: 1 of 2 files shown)


hgaiser:
Thanks!

Issue #131, opened 2026-07-23
flatpak steam launches on main screen and audio, not the dedicated XWayland


niv:
## Issue

All discovered Steam games (and steam itself) launches on the main desktop screen, not in the background. Audio plays through the desktop speakers, not the target devices.

This issue is specific to flatpak'ed Steam. Other flatpak apps and games seem to work.

## Version

TOT compiled manually (0.13.5 as of a few minutes ago), launched in a terminal via `cargo run --release --  ~/.config/moonshine/config.toml`.

## Additional testing done (see config below)

- glxgears works but is cut off except the top left quadrant. Not bothered about it.
- Freelancer HD and bsnes works: both bottles.app and other "native" flatpak apps work. Both video and audio plays correctly on target device.
- Native steam was not tested; I prefer flatpak for "sandboxing".

## Configuration extract

```
[[application]]
title = "bsnes"
command = [ "flatpak", "run", "dev.bsnes.bsnes" ]
launch_timeout_secs = 2

[[application]]
title = "Freelancer"
command = [ "flatpak", "run", "--command=bottles-cli", "com.usebottles.bottles",
    "run", "-b", "Freelancer", "-p", "Freelancer HD" ]
launch_timeout_secs = 2

[[application]]
title = "glxgears"
command = [ "/usr/bin/glxgears" ]
launch_timeout_secs = 2

[[application]]
title = "Steam"
command = [ "flatpak", "run", "com.valvesoftware.Steam", "steam://open/bigpicture" ]
launch_timeout_secs = 2

[[application_scanner]]
type = "steam"
library = "$HOME/.var/app/com.valvesoftware.Steam/.local/share/Steam"
command = [ "flatpak", "run", "com.valvesoftware.Steam", "steam://rungameid/{game_id}" ]
launch_timeout_secs = 2
```

## More

Thank you for your work on this. It'll be a perfect fit for my environ (caught the post on HN to learn about it).

niv:
I found a workaround/fix:

```
[[application_scanner]]
type = "steam"
library = "$HOME/.var/app/com.valvesoftware.Steam/.local/share/Steam"
command = [ "dbus-run-session", "--",
    "flatpak", "run", "com.valvesoftware.Steam",
    "-newbigpicture", "steam://rungameid/{game_id}" ]
```

The notable addition is the dbus-run-session -- prefix. For whatever reason, this now summons Steam and it's games on the streaming target.

[...]

PR #152, opened 2026-07-30
feat: add Heroic Games Launcher application scanner

### LINKED ISSUE #149: [Feature request] Heroic Games Launcher application scanner
Moonshine auto-discovers games from Steam and Lutris via `application_scanner`; it would be nice to have the same for [Heroic Games Launcher](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher). It covers Epic, GOG, and Amazon libraries.

scottjab:
Adds a scanner for [Heroic Games Launcher](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher),
so Epic, GOG and Amazon games get picked up the same way Steam and Lutris do.

## Usage

```toml
[[application_scanner]]
type = "heroic"
command = ["heroic", "--no-gui", "heroic://launch?appName={app_name}&runner={runner}"]
```

`{app_name}` is Heroic's internal game id and `{runner}` is the store it came
from (`legendary`, `gog`, `nile`, or `sideload`). Commands go through PATH so a
bare `heroic` works on most distros. The README also covers the `flatpak run`
form.

`config_dir` defaults to `$XDG_CONFIG_HOME/heroic` and falls back to the Flatpak
location if that's the only one installed.

## How it works

Heroic keeps a library cache per backend under its config dir. The scanner reads
all four, keeps whatever is flagged installed, and skips DLC. The DLC bit
matters more than it sounds: Heroic lists installed DLC as its own library
entry, so Hitman 3 by itself turned into 20 apps before filtering.

The backends can't agree on whether the game array is keyed `library` or
`games`, so both work. Entries are parsed one at a time, so a single weird game
doesn't take out the rest of the library. Box art comes from whatever Heroic has
already downloaded.

## Testing

Ran it against a real Heroic install (native, Epic, 47 games, 2 installed). Gets
exactly the two installed games, DLC filtered out, portrait art on both. 9 unit
tests, fmt and clippy clean.

Stuff I couldn't check:

- **GOG and Amazon.** I only have an Epic library, so those paths are built from
  Heroic's shared game schema rather than anything I could observe. The
  `library`/`games` key handling is the most likely thing to need a fix.
- **Flatpak.** The Flatpak config path comes from Flatpak's standard layout, I
  didn't test it on an actual Flatpak install.

Closes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,32 @@ Box art is automatically loaded from Lutris's `coverart/` directory when availab
 The default database path is `~/.local/share/lutris/pga.db`.
 You can override it with the `pga_db` option to point at a custom database location.
 
+**Heroic scanner** — finds all installed [Heroic Games Launcher](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher) games (Epic, GOG, Amazon and manually added games):
+
+```toml
+[[application_scanner]]
+type = "heroic"
+command = ["heroic", "--no-gui", "heroic://launch?appName={app_name}&runner={runner}"]
+```
+
+If you installed Heroic from Flathub, launch it through Flatpak instead:
+
+```toml
+[[application_scanner]]
+type = "heroic"
+command = [
+  "flatpak", "run", "com.heroicgameslauncher.hgl",
+  "--no-gui", "heroic://launch?appName={app_name}&runner={runner}",
+]
+```
+
+`{app_name}` is Heroic's internal game id and `{runner}` is the store it came from (`legendary` for Epic, `gog` for GOG, `nile` for Amazon, `sideload` for manually added games).
+Commands are looked up in `PATH`, so a bare `heroic` works regardless of where your distribution installs it; use an absolute path if you run Heroic from an AppImage.
+Box art is automatically loaded from Heroic's `images-cache/` directory for any art Heroic has already downloaded.
+
+The default configuration directory is `~/.config/heroic`, falling back to `~/.var/app/com.heroicgameslauncher.hgl/config/heroic` when only the Flatpak is installed.
+You can override it with the `config_dir` option.
+
 ## Tips & Tricks
 
 See [TIPS.md](TIPS.md) for practical recipes and workarounds.
--- a/README.md
+++ b/README.md
@@ -254,13 +254,18 @@ If you installed Heroic from Flathub, launch it through Flatpak instead:
 [[application_scanner]]
 type = "heroic"
 command = [
+  "dbus-run-session", "--",
   "flatpak", "run", "com.heroicgameslauncher.hgl",
   "--no-gui", "heroic://launch?appName={app_name}&runner={runner}",
 ]
 ```
 
+The `dbus-run-session` prefix keeps the game inside Moonshine's compositor instead of on your desktop.
+See [TIPS.md](TIPS.md) for why Flatpak needs it.
+
 `{app_name}` is Heroic's internal game id and `{runner}` is the store it came from (`legendary` for Epic, `gog` for GOG, `nile` for Amazon, `sideload` for manually added games).
-Commands are looked up in `PATH`, so a bare `heroic` works regardless of where your distribution installs it; use an absolute path if you run Heroic from an AppImage.
+Commands are resolved against Moonshine's own `PATH`, which under the packaged systemd service is systemd's default rather than your login shell's.
+A bare `heroic` therefore works for packages that install into `/usr/bin`, but not for an AppImage or anything under `~/.local/bin`; use an absolute path in those cases.
 Box art is automatically loaded from Heroic's `images-cache/` directory for any art Heroic has already downloaded.
 
 The default configuration directory is `~/.config/heroic`, falling back to `~/.var/app/com.heroicgameslauncher.hgl/config/heroic` when only the Flatpak is installed.
--- a/moonshine-core/src/app_scanner/heroic.rs
+++ b/moonshine-core/src/app_scanner/heroic.rs
@@ -75,10 +75,8 @@ fn default_config_dir() -> PathBuf {
 /// the array lives under, so both spellings are accepted.
 #[derive(Debug, Deserialize)]
 struct HeroicLibrary {
-	/// Left untyped so a single malformed game can be dropped without losing
-	/// the whole library.
 	#[serde(alias = "games", default)]
-	library: Vec<serde_json::Value>,
+	library: Vec<HeroicGame>,
 }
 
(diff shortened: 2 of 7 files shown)


moonshine-core/src/app_scanner/heroic.rs
```diff
@@ -0,0 +1,490 @@
...
+	}
+}
+
+/// The contents of a Heroic library cache.
+///
+/// Every backend serializes the same game shape, but they disagree on the key
+/// the array lives under, so both spellings are accepted.
+#[derive(Debug, Deserialize)]
+struct HeroicLibrary {
+	/// Left untyped so a single malformed game can be dropped without losing
```
hgaiser: Did you actually run into malformed games? Why were they malformed?

moonshine-core/src/app_scanner/heroic.rs
```diff
@@ -0,0 +1,490 @@
...
+	let config_dir = &config.config_dir;
+
+	let binding = config_dir.to_string_lossy();
+	let expanded = shellexpand::full(&binding)
+		.map_err(|e| tracing::warn!("Failed to expand Heroic configuration path {:?}: {e}", config_dir))?;
+
+	let config_dir = PathBuf::from(expanded.as_ref());
+
+	if !config_dir.exists() {
+		tracing::debug!("Heroic configuration directory not found at {:?}.", config_dir);
```
hgaiser: Hmm I am thinking it makes more sense if this is a warning. If the user configured a Heroic scanner but it doesn't run, it should be warned.

README.md
```diff
@@ -240,6 +240,32 @@ Box art is automatically loaded from Lutris's `coverart/` directory when availab
...
+[[application_scanner]]
+type = "heroic"
+command = [
+  "flatpak", "run", "com.heroicgameslauncher.hgl",
+  "--no-gui", "heroic://launch?appName={app_name}&runner={runner}",
+]
+```
+
+`{app_name}` is Heroic's internal game id and `{runner}` is the store it came from (`legendary` for Epic, `gog` for GOG, `nile` for Amazon, `sideload` for manually added games).
+Commands are looked up in `PATH`, so a bare `heroic` works regardless of where your distribution installs it; use an absolute path if you run Heroic from an AppImage.
```
hgaiser: I'm also not sure what PATH is exactly when moonshine is running through a systems service. It might not have picked up Heroic in that scenario?

README.md
```diff
@@ -240,6 +240,32 @@ Box art is automatically loaded from Lutris's `coverart/` directory when availab
...
 
+**Heroic scanner** — finds all installed [Heroic Games Launcher](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher) games (Epic, GOG, Amazon and manually added games):
+
+```toml
+[[application_scanner]]
+type = "heroic"
+command = ["heroic", "--no-gui", "heroic://launch?appName={app_name}&runner={runner}"]
+```
+
+If you installed Heroic from Flathub, launch it through Flatpak instead:
```
hgaiser, line 251: This might be worth it to test, because there were some issues launching Steam through Flatpak earlier: https://github.com/hgaiser/moonshine/issues/131

hgaiser (commented):
Thanks! Left some small comments.

Reading up on this I found https://github.com/Rolv-Apneseth/lib_game_detector , which might be be interesting. Unfortunately they use a non-permissive license (AGPLv3.0). I [asked](https://github.com/Rolv-Apneseth/lib_game_detector/issues/56) them if they are open to dual licensing.

moonshine-core/src/app_scanner/heroic.rs
scottjab: No, I didn't. That was me being defensive about a problem I never actually saw, so I've dropped it. `HeroicLibrary.library` is `Vec<HeroicGame>` now and parses straight through; a bad entry fails the library like any other parse error and still logs at warn. Removed the matching test with it.

moonshine-core/src/app_scanner/heroic.rs
scottjab: Good call, bumped it to warn and spelled out the consequence:

```rust
tracing::warn!(
    "Heroic configuration directory not found at {:?}, no Heroic games will be scanned.",
    config_dir
);
```

Worth noting `lutris.rs:60` does the same thing at debug for a missing `pga.db`. Same reasoning applies there. Want me to fix it in this PR or leave it for a separate one?

README.md
scottjab: You're right to poke at this, the line is too breezy. `build_exec_entry` (`session/application.rs:696`) resolves with `which::which` inside moonshine's own process, so it's moonshine's PATH that matters, not systemd's exec search path. And `moonshine@.service` only sets `MOONSHINE_LOG`, so under the packaged service that PATH is systemd's built-in default, roughly `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`.

So a bare `heroic` resolves for the .deb/.rpm/AUR installs that drop a symlink in `/usr/bin`, but not for anything under `~/.local/bin` or `~/Applications`. Those will work fine when you test in a terminal and then fail under the service, which is the worst version of this. Reworded to say that instead of "regardless of where your distribution installs it".

README.md
scottjab: I don't currently have a good machine to test Flatpak on, so I haven't verified this.

Reasoning from your TIPS.md writeup though, the root cause is Flatpak's portal handing back the host Wayland socket, which isn't Steam specific, so Heroic through Flatpak should hit the same thing. The example I added has no `dbus-run-session --` prefix, so I've added it.

Worth flagging separately: Heroic is Electron and `heroic://launch?...` goes through its URL handler, so if Heroic is already running on the desktop the new invocation may hand the URL to that instance and the game opens on the desktop instead. That would affect the native config too. I can test that one, it doesn't need Flatpak.

scottjab:
Library looks good. If they go for dual licensing, a refactor onto it would be a good idea.

moonshine-core/src/app_scanner/heroic.rs
hgaiser: Ah yeah might as well fix it in this PR 👍 . Thanks.

README.md
hgaiser: Yeah same thing more or less applies to Steam I guess. There's a tip to close Steam on the host. I don't think we need the same for Heroic, users should be able to manage I hope 🤞

moonshine-core/src/app_scanner/heroic.rs
scottjab: Fixed lutris in 27bc41e.

PR #154, opened 2026-08-01
fix(hdr): make the injected AV1 HDR metadata OBUs spec-valid

lutyjj:
Found by Fable when investigating a green screen in the AV1 HDR path during AC4 Resynced gameplay. It turned out not to be THE reason for that green screen (that one is the SDR->HDR switch mid-stream, https://github.com/hgaiser/moonshine/issues/155), but it's an actual bug on its own. Fable is much smarter than me, so I had to go study a little how things operate under the hood in AV1 - attaching references for future folks that might land on this PR:

* [AV1 spec, 5.3.1 General OBU syntax](https://aomediacodec.github.io/av1-spec/#general-obu-syntax) - every OBU except OBU_TILE_GROUP, OBU_TILE_LIST and OBU_FRAME ends with `trailing_bits()`, and `obu_size` counts them
* [AV1 spec, 6.7.4 Metadata high dynamic range mastering display color volume semantics](https://aomediacodec.github.io/av1-spec/#metadata-high-dynamic-range-mastering-display-color-volume-semantics) - chromaticities are `x/65536`, `luminance_max` is in 1/256 cd/m² steps, `luminance_min` in 1/16384
* [H.265 spec](https://www.itu.int/rec/T-REC-H.265), D.3.28 mastering display colour volume SEI - the units `HdrMetadata` carries (0.00002 chromaticity steps, 0.0001 cd/m² luminance), which the H.26x SEI path writes as-is
* [dav1d src/obu.c, check_trailing_bits](https://code.videolan.org/videolan/dav1d/-/blob/master/src/obu.c) - the real-world enforcement: no trailing bits, no frame

The AV1 links point at the HTML working copy since it's the only deep-linkable form; the approved spec is the [PDF](https://aomediacodec.github.io/av1-spec/av1-spec.pdf), same section numbers.

With that out of the gate, this PR makes the HDR metadata OBUs we inject before AV1 key frames spec-valid:

1. Terminates the CLL and MDCV OBUs with trailing bits, counted in `obu_size`. Without them dav1d rejects the OBU and loses the key frame it precedes with it - in our tests a pre-fix stream decodes to exactly zero frames.
2. Converts MDCV values from the SEI units into AV1's units when building the OBU. Decoders accepted them before, but read nonsense: our 0.680 red primary decoded as 0.540, 1000 nits max luminance as 39062. Wrong tone mapping, not a decode failure.

Testing:

* End to end on real NVENC AV1 output (4K, 10-bit, full range, mid-stream SDR<->HDR switches): pre-fix streams hard-fail in dav1d ("Invalid argument", zero frames) and ffmpeg's parser; post-fix, dav1d 1.5.4, ffmpeg and nvdec decode every frame, and ffprobe reads the exact metadata values back.
* Also decodes clean on a Steam Machine's VAAPI stack.

### DIFF
--- a/moonshine-core/src/session/stream/video/pipeline/hdr_sei.rs
+++ b/moonshine-core/src/session/stream/video/pipeline/hdr_sei.rs
@@ -138,7 +138,7 @@ fn build_av1_cll_obu(m: &HdrMetadata) -> Vec<u8> {
 	payload.extend(m.max_cll.to_be_bytes());
 	payload.extend(m.max_fall.to_be_bytes());
 
-	build_av1_obu(5, &payload) // obu_type = OBU_METADATA = 5
+	build_av1_metadata_obu(&payload)
 }
 
 /// Build AV1 OBU metadata for HDR MDCV (metadata_type = 2).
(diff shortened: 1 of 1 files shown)


hgaiser:
Honestly, my knowledge of these specifics is quite limited as well. The fix looks good though, so thanks 👍

PR #157, opened 2026-08-02
fix(video): split GSO sends to fit the kernel's UDP limits

### LINKED ISSUE #156: [Bug] GSO send path loses video data at high bitrates, stream freezes and never recovers (potentially only occurs on VMs)
I was testing latest moonshine-git at 4K/120Hz/150Mbps on my setup and the stream consistently freezes within seconds and never recovers - on AV1 it additionally ends in a greenscreen (probably related to #155). HEVC freezes the same way, so it's not codec-specific. 1080p/28Mbps is working fine on the exact same machines.

- Host: NVidia 5060 Ti (driver 610.43.03), Arch, latest moonshine-git. It's a VM on Proxmox with GPU passthrough - that may be a precondition, see below.
- Clients: SteamMachine (moonlight-qt from flathub) and a CachyOS machine with an RTX 4090. SteamMachine over wired, Cachy PC over WiFi. Both reproduce the same freeze, so it's not a client thing either.

First thought something's up with my network: a sustained 150Mbps UDP burst host→client over plain `sendto` gives zero send errors and zero drops on both ends, and the stream itself holds exactly 150Mbps flawlessly right up until the collapse.

Started bisecting recent commits which pointed at the GSO path from #138, confirmed two ways:

1. 85ebb8c (pre-#138) streams 4K/120/150 perfectly; 422d56b (#138 merged) freezes.
2. Current main with only GSO forced off is perfectly smooth again - same machines, same session:

```rust
// session/stream/video/mod.rs - both gso checks
let gso_enabled = false; // instead of udp_state.max_gso_segments() > 1
let use_gso = false;
```

So with GSO on, the send path (quinn-udp 0.6.1, `UDP_SEGMENT`) loses video data at high bitrate somewhere between the socket and the wire. The freeze and greenscreen are downstream of that loss: the client asks for recovery via RFI/IDR, and on AV1 the recovery key frames never decode cleanly (again, likely #155, have potential fix for it already).

What I haven't pinned down yet is where exactly the segments die. If this wasn't spotted before, I assume that my setup might be unique since it's a VM. Bare-metal hosts may never see this - would love confirmations either way from anyone streaming >100Mbps with a GSO on.

lutyjj:
Closes #156.

The GSO path handed each frame's entire shard batch to a single send - 150-350KB at 4K bitrates. The kernel rejects `UDP_SEGMENT` sends over 65507 bytes with EMSGSIZE before anything reaches the wire, and quinn-udp's `send()` masks every error except `WouldBlock` as `Ok`, so oversized frames vanished silently and the per-shard fallback was dead code. At 150Mbps that's practically every frame; 28Mbps frames (~29KB) fit, which is why low bitrates were fine. Easy to check on loopback: 46 segments x 1404B deliver as 46 packets, 47 segments fail with EMSGSIZE.

The fix:

1. Split each batch into chunks of `min(max_gso_segments, 65507 / shard_size)` segments - a 250-shard frame goes out in 6 syscalls instead of 250, keeping the #138 win/s.
2. Use `try_send` so kernel errors actually reach the per-shard fallback; the warning is rate-limited to once per second.
3. Drain `WouldBlock` by awaiting writability and resending the same chunk - a frame burst can outrun the socket buffer.
4. Wrap sends in `wrap_cancel` so a stalled socket can't block session shutdown.
5. Unit tests for the cap arithmetic and chunk invariants. `ShardBatch::shards()` lost its last caller and is removed.

Testing: `cargo test -p moonshine-core` for the new tests; end to end at 4K/120Hz/150Mbps AV1 and HEVC - pre-fix freezes within seconds, post-fix streams flawlessly with GSO active and zero fallback warnings. 1080p/28Mbps unchanged.

---
Host: NVidia 5060 Ti (driver 610.43.03), Arch, VM on Proxmox with GPU passthrough
Client: SteamMachine (moonlight-qt from flathub), wired gigabit

### DIFF
--- a/moonshine-core/src/session/stream/video/gso_socket.rs
+++ b/moonshine-core/src/session/stream/video/gso_socket.rs
@@ -1,0 +1,171 @@
+use quinn_udp::{Transmit, UdpSockRef, UdpSocketState};
+use tokio::net::UdpSocket;
+
+use super::shard_batch::ShardBatch;
+
+/// Maximum payload of one UDP datagram (65535 minus IPv4/UDP headers).
+/// A GSO batch is still one datagram: the kernel rejects anything larger
+/// with EMSGSIZE before segmenting. IPv6 allows 20 more bytes; the IPv4
+/// value is safe for both.
+const MAX_UDP_PAYLOAD: usize = 65507;
+
+/// Number of shards per GSO send: the kernel's segment-count cap or the
+/// datagram-size cap, whichever binds first. Never zero, even for a shard
+/// larger than a datagram (such a chunk fails at send time and falls back
+/// to per-shard sends, which fail visibly instead of silently).
+fn gso_segments_per_send(max_gso_segments: usize, shard_size: usize) -> usize {
+	max_gso_segments
+		.min(MAX_UDP_PAYLOAD.checked_div(shard_size).unwrap_or(0))
+		.max(1)
+}
+
+/// UDP socket for the video stream, with Generic Segmentation Offload.
+///
+/// Wraps a `tokio::net::UdpSocket` and quinn-udp's `UdpSocketState` so shard
+/// batches can be sent as GSO super-packets sized to the kernel's segment-count
+/// and datagram-size caps. Batches exceeding a cap are split into chunks; chunks
+/// the kernel rejects fall back to per-shard sends. Without GSO support, the
+/// socket sends everything per-shard.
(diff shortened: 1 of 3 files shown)


hgaiser:
Hmm I guess my tests were using smaller packets, nice catch!

I moved the GSO socket into its own struct but kept the logic the same. Thanks for looking into this!

PR #167, opened 2026-08-05
fix(audio): tolerate timerfd wakeups with no expirations

lutyjj:
Had a live session die mid-game with `Session stopped unexpectedly (reason: PulseServerStopped)` - the audio clock's timerfd read returned EAGAIN and the `?` propagated it as fatal, tearing down the whole session (and the running app with it). A wakeup can race its own drain and find no expirations left to read; hit it in the wild under host memory pressure where the event loop was stalling (12s of buffer-underrun warnings first, then the EAGAIN).

The fix skips the tick on WouldBlock instead, which is how the accept and client-read arms of the same event loop already treat it. Real errors still propagate.

```
WARN  ...pulse_server: Buffer underrun for stream 6   (x many, ~12s)
ERROR ...pulse_server: PulseServer error: IO error: Resource temporarily unavailable (os error 11)
WARN  ...manager: Session stopped unexpectedly (reason: PulseServerStopped), waiting for new session.
```

---
Host: NVidia 5060 Ti (driver 610.43.03), Arch

### DIFF
--- a/moonshine-core/src/session/stream/audio/pulse_server/mod.rs
+++ b/moonshine-core/src/session/stream/audio/pulse_server/mod.rs
@@ -315,8 +315,13 @@ impl PulseServer {
 			for event in events.iter() {
 				match event.token() {
 					CLOCK => {
-						self.clock.read()?;
-						self.clock_tick()?;
+						// A wakeup can race its own drain and find no expirations
+						// left to read (EAGAIN); skip the tick, don't kill the session.
+						match self.clock.read() {
+							Ok(_) => self.clock_tick()?,
+							Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => (),
+							Err(e) => return Err(e.into()),
+						}
 					},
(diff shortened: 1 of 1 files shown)


lutyjj:
keeping this as draft for now - the diagnosis is from logs of a single real crash and the fix follows the loop's existing WouldBlock convention, but I haven't reproduced the race to verify it actually prevents the teardown. it's running on my host now; will un-draft once it survives a few sessions under the same memory-pressure conditions (or I manage to repro it synthetically).

angusburroughs:
Encountering exactly the same crashes. @lutyjj did your testing of the fix go well?

lutyjj:
@angusburroughs yeah, tbh I forgot about this PR, I was running with it ever since I submitted it just fine :)

angusburroughs:
Fantastic, in that case are you ready to un-draft it?

hgaiser:
👍 thanks

PR #169, opened 2026-08-07
fix: discover Heroic libraries and read GOG install state from its manifest

### LINKED ISSUE #160: Heroic scanner only picks up one "store"
I have games installed from GoG and some are sideloaded. The Heroic scanner only picks up the sideloaded ones.

Both my `store_cache/gog_library.json` and `sideload_apps/library.json` files are populated.

Here is the relevant part of my nixconfig:

```nix
# Everything from the Configuration section of the main README goes
# here, written as nix instead of TOML.
settings = {
  application = [
    {
      title = "Steam Big Picture";
      command = [
        "/run/current-system/sw/bin/steam"
        "steam://open/bigpicture"
      ];
    }
  ];
  application_scanner = [
    {
      type = "steam";
      library = "$HOME/.local/share/Steam";
      command = [
        "/run/current-system/sw/bin/steam"
        "-bigpicture"
        "steam://rungameid/{game_id}"
      ];
    }
    {
      type = "heroic";
      command = [
        "/run/current-system/sw/bin/heroic"
        "--no-gui"
        "heroic://launch?appName={app_name}&runner={runner}"
      ];
    }
  ];
};
```

scottjab:
The Heroic scanner never returns GOG games, which is why #160 looked like only one store works.

## GOG install state

Heroic merges install state into a *copy* of each GOG game and keeps that copy out of the library cache it writes to disk, in `storeManagers/gog/library.ts`:

```js
// Create new object to not write install data into library store
const copyObject = Object.assign({}, unifiedObject)
if (installedInfo) {
  copyObject.is_installed = true
  copyObject.install = installedInfo
}
library.set(copyObject.app_name, copyObject)   // in-memory only
...
libraryStore.set('games', gamesObjects)        // disk, without install state
```

`gogToUnifiedInfo` hardcodes `is_installed: false`, so every entry in `store_cache/gog_library.json` claims to be uninstalled and the scanner drops the lot. Epic, Amazon and sideloaded entries record install state in their own caches, so only GOG is affected. Same behavior back through v2.14.0, so it isn't a recent regression.

Fall back to `<runner>_store/installed.json`, keyed by `appName` to match the library entry's `app_name`. Heroic rewrites it on every install and uninstall rather than only on a full library refresh, so it's fresher than the cache would have been either way. A missing manifest reads as empty, so stores that report `is_installed` honestly are unaffected.

The library flag is still honored alongside the manifest. That keeps the synthetic `gog-redist` entry, the one GOG game Heroic always marks installed, going through the existing DLC filter that already hides "Galaxy Common Redistributables".

## Library discovery

The four library paths were hardcoded. They're now discovered by suffix under `store_cache/`, plus `sideload_apps/library.json`. That directory holds plenty besides libraries (install info, achievements, cached API responses), so only `*_library.json` is matched.

This picks up **Zoom Platform**, the fifth backend Heroic shipped in v2.22, which wasn't being scanned at all. A sixth needs no change here. The runner already comes from the library entry, so the installed manifest path follows from it rather than needing its own table.

## Box art

`icons/<app_name>.png` was being preferred over the real cover art. `shortcuts/utils.ts::getIcon` writes the portrait art it downloads as `.jpg` for every store, and only writes a `.png` when it has substituted a store's own icon to keep transparency:

```js
if (productApiData && productApiData.data.images?.icon) {
  image = 'https:' + productApiData.data.images?.icon
  icon = `${iconsFolder}/${appName}.png` // Allow transparency
}
```

That icon is a square logo, not cover art. GOG is the only backend it happens for today, so matching on the extension rather than the runner covers any other it starts happening for. GOG art is cached under the unmodified `art_square` URL, since `getImageFormatting` only appends the resizing parameters for Epic, so the existing image cache lookup finds it.

## Testing

Unit tests cover each of the above. Also checked against a real Heroic config with 47 Epic entries: discovery finds the three caches and the scan returns the 3 installed non-DLC games with their `icons/*.jpg` art, unchanged from before.

I don't own any GOG, Amazon or Zoom games, so those paths are verified against Heroic's source rather than a live library. @istvnurbn confirmed the GOG fix works on their setup, and reported the box art problem that the last section fixes.

Fixes #160

### DIFF
--- a/moonshine-core/src/app_scanner/heroic.rs
+++ b/moonshine-core/src/app_scanner/heroic.rs
@@ -65,7 +65,7 @@ fn find_libraries(config_dir: &Path) -> Vec<PathBuf> {
 			})
 			.collect(),
 		Err(e) => {
-			tracing::debug!("Failed to read Heroic store cache {:?}: {e}", store_cache);
+			tracing::warn!("Failed to read Heroic store cache {:?}: {e}", store_cache);
 			Vec::new()
 		},
 	};
--- a/moonshine-core/src/app_scanner/heroic.rs
+++ b/moonshine-core/src/app_scanner/heroic.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


istvnurbn:
@scottjab It works, but with one caveat. For GoG games, it displays the icons instead of the actual cover art ('art_square' in the JSON) in Moonlight. But this solves the main issue! Thanks for your prompt response!

scottjab:
@istvnurbn Try again I think the current commit should work.  You might have to clear the cache in your moonlight client though.

istvnurbn:
@scottjab works flawlessly, you are a star!

scottjab:
Incredible, thanks for testing!

moonshine-core/src/app_scanner/heroic.rs
```diff
@@ -35,14 +36,65 @@ pub struct HeroicApplicationScannerConfig {
...
+			.flatten()
+			.map(|entry| entry.path())
+			.filter(|path| {
+				path.file_name()
+					.and_then(|name| name.to_str())
+					.is_some_and(|name| name.ends_with(LIBRARY_SUFFIX))
+			})
+			.collect(),
+		Err(e) => {
+			tracing::debug!("Failed to read Heroic store cache {:?}: {e}", store_cache);
```
hgaiser: This should be a warning. If the user configured Heroic parsing then we should warn if it fails.

moonshine-core/src/app_scanner/heroic.rs
```diff
@@ -221,13 +329,15 @@ fn is_valid_app_name(app_name: &str) -> bool {
 /// only gets used when there is nothing better.
 fn find_boxart(config_dir: &Path, game: &HeroicGame) -> Option<PathBuf> {
 	// Full resolution portrait art, saved when Heroic downloads a game icon.
+	//
+	// Heroic always writes that art as a JPG. It only writes a PNG when it has
+	// substituted a store's own icon for the art, to keep the transparency, and
+	// that icon is a small square logo rather than cover art. Matching on the
+	// extension alone keeps any store Heroic does that for out of the box art.
```
hgaiser: This feels a bit hacky, is there no better way to find boxart?

hgaiser (commented):
Thanks, two minor questions, but LGTM overall.

moonshine-core/src/app_scanner/heroic.rs
scottjab: Yeah, fair. Went and read Heroic's source.

`icons/` is written by the shortcuts code, `getIcon()` in `shortcuts/utils.ts`, and the only callers are "create desktop shortcut" and "add to Steam". So it's not an art cache, a file only lands there if you made a shortcut for that game. `images-cache` is the one that fills in whenever the UI draws anything.

So I flipped the order. Cache first, icons as fallback, extension check gone. The GOG logo only turns up when there's nothing else now, which is fine for a fallback.

Checked my own library and every game with a shortcut icon also had cached art, so the reorder shouldn't cost anything in practice.

Also means it's just a reorder, main already accepted both jpg and png there.

hgaiser:
Thanks!

PR #170, opened 2026-08-11
fix: use fail mode for Plasma sessions to avoid systemd reload conflict

brankosimic:
When streaming a KDE Plasma desktop session, the Plasma session startup fails with:

> Could not start systemd managed Plasma session: "org.freedesktop.systemd1.TransactionIsDestructive"

## What happened

The journal logs show the sequence:

1. `moonshine` calls `StartTransientUnit("moonshine-session.service", "replace", ...)`
2. The `"replace"` mode triggers a systemd reload of the transient unit
3. Plasma's systemd integration (which monitors the transient unit it runs in) interprets the reload as a session teardown signal
4. Plasma tries to tear down `graphical-session.target` while simultaneously starting `plasma-workspace-wayland.target`
5. systemd rejects this as a destructive transaction conflict

## Why this fix works

Plasma's systemd integration reacts to the reload triggered by `"replace"` mode. Using `"fail"` mode avoids the reload entirely — it simply fails if the unit already exists. Since `stop_unit` is always called beforehand to clean up leftover units, `"fail"` mode never actually fails in practice.

The fix is conditional: only Plasma desktop sessions (`startplasma-way`, `startplasma-wayland`, `plasmashell`) use `"fail"` mode. All other sessions (games, Steam, etc.) continue using `"replace"` with no behavior change.

## Related: SinglePixelBufferState

Added `SinglePixelBufferState` to the compositor state. This enables the single-pixel buffer protocol for Wayland clients that require it (e.g. certain GTK4 UI elements).

## Testing

- All CI checks pass: fmt, clippy, tests (93), docs, machete
- Tested with KDE Plasma desktop stream — no more transaction conflict

### DIFF
--- a/moonshine-core/src/session/application.rs
+++ b/moonshine-core/src/session/application.rs
@@ -109,6 +109,7 @@ pub(crate) struct LaunchOptions<'a> {
 	pub post_commands: &'a Vec<Vec<String>>,
 	pub stdout_value: &'a Option<String>,
 	pub stderr_value: &'a Option<String>,
+	pub job_mode: &'static str,
 }
 
 /// Runtime context required to launch an application.
@@ -157,6 +158,8 @@ impl Application {
 		// Stop any leftover unit from a previous session.
 		let _ = stop_unit(&conn, &context.unit_name).await;
 
+		let job_mode = if is_plasma_session(program) { "fail" } else { "replace" };
+
 		// Launch the application as a transient systemd service unit.
 		let options = LaunchOptions {
 			unit_name: &context.unit_name,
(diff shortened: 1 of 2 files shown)


hgaiser:
Hey, curious how you resolved this issue?

brankosimic:
I resolved most of it, but it was still affecting my main KDE desktop. In order to fix, had to run plasma sessions in their own private systemd user manager + private D-Bus. Did not push that as I did not like it, so dropped it.
Then decided to do something else which for my use case is even better. That is to stream the current user desktop like sunshine does. It is a bit different from the project vision, but could come in handy. I am still testing that.

PR #172, opened 2026-08-11
update nix section with newly added module and pkg

tree-sapii:
https://github.com/NixOS/nixpkgs/pull/544596 got merged into nixpkgs last month with the the pkg already existing in unstable [PR tracker](https://nixpk.gs/pr-tracker.html?pr=544596).  
There is also a module init pr https://github.com/NixOS/nixpkgs/pull/544393, however its still not merged. I will open this as draft until that is merged.

I wonder if its best to leave the existing instructions on how to use this project's consumable module or to delete them and the module file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,16 @@ nix profile install github:hgaiser/moonshine
 
 After installing, follow the [Enable the service](#enable-the-service) steps below (the NixOS module handles this for you when enabled).
 
+Moonshine is now directly packaged in nixpkgs, you can add it to your system packages or enable the module directly:
+
+```nix
+environment.systemPackages = with pkgs; [ moonshine ];
+services.moonshine = {
+	enable = true;
+	# Checkout more settings at https://search.nixos.org/options?query=moonshine
+};
+```
+
 ### Bazzite / Silverblue / Atomic distros
 
 On Fedora Atomic distros, download the `.rpm` from the [releases page](https://github.com/hgaiser/moonshine/releases) and layer it:


hgaiser:
I'm not that familiar with the NixOS ecosystem. My philosophy regarding moonshine is generally, if there is a preferred approach, let's not clutter things with other approaches.

Am I correct that if https://github.com/NixOS/nixpkgs/pull/544393 is merged, you can add `moonshine` to the list of installed packages in your config and that's it? Is there still a reason to host nix files in this repository if that gets merged?

repparw:
package's already merged from another pr, that pr remains with the module implementation. 

ideally you could keep the flake for dev and qa of the package (in nix you can use this repo as an input, that's what the current setup does to use the repo defined module and package)

you could drop the module from the repo though when/if it gets merged, as you can just override the module's package option for testing the flake one, without replacing the canonical module implementation from nixpkgs

if it's going to be 'blessed' by you upstream, you might want to say that there (they might add you as maintainer for the package, so eg you can approve updates when you make a release if another maintainer doesn't see it before you)

tree-sapii:
You could also copy nixpkgs's module in here for the dev module to not have inconsistencies with what most users are going to use

repparw:
sure, once it lands it could be easier to keep both things. i would wait until it gets approval as it seems to have a couple things that still need ironing out compared to the shipped one here

tree-sapii:
https://github.com/NixOS/nixpkgs/pull/544393 merged, will update soon.

PR #173, opened 2026-08-11
fix: dont name an unknown pa source

KarlVogel:
Specifying a default_source_name causes issues for apps that will then try to use that source, since there is no server implementation for audio sources yet.

Caused the Project Zomboid native linux client to crash in libfmod.so


hgaiser:
Looks good to me, thanks!

PR #175, opened 2026-08-14
feat: add native touch and pen input

antomanc:
## Summary

- advertise Moonlight's native pen/touch capability alongside controller touchpad support
- parse native multi-touch and pen packets from the control stream
- expose a Wayland touch device and tablet tool with pressure, hover distance, tilt, eraser, and stylus-button support
- ignore null/unmapped keyboard packets emitted by some touch clients and correctly decode 16-bit VK codes
- document touchscreen and pen/stylus support

## Implementation notes

- Touch contacts use their Moonlight pointer IDs as stable Wayland touch slots.
- Coordinates are clamped and mapped from normalized client coordinates into the active output.
- Pen cancel and hover-leave events release tip, buttons, and proximity state.
- Malformed, unknown, and non-finite pointer packets are rejected before reaching the compositor.

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` (113 tests passed)
- `cargo build --release --workspace`
- `git diff --check`
- connected a Moonlight touch client over LAN, launched Steam at 2778x1940/120 Hz with H.265, and verified native touch input without keyboard-warning spam

The installed AUR package was left unchanged; live validation used the release binary built from this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Your keyboard, mouse, and controller inputs are sent back to the host so you can
 - **No monitor required**: Works on headless servers — no HDMI dummy plug needed.
 - **Hardware video encoding**: H.264, H.265, and AV1 encoding using the GPU.
 - **HDR support**: True 10-bit HDR streaming for supported games.
-- **Full input support**: Mouse, keyboard, and gamepad (including motion, touchpad, and haptics).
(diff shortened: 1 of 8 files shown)


hgaiser:
Thanks!

PR #187, opened 2026-08-28
fix(nix): disable nixpkgs module

amateurattentionseeker:
Rebuilding NixOS causes a failure because the latest `nixos-unstable` added their own separate moonshine module conflicting with the flake module. This PR fixes it by disabling the nixpkgs module.


hgaiser:
@scottjab ? 🙄

scottjab:
Lgtm

PR #190, opened 2026-08-29
fix(nix): fix PR 187

amateurattentionseeker:
I apologise for inconvenience and dealing with this. Somehow the build failure slipped under my radar when I was testing this. This should actually fix it now


hgaiser:
I will have to take your word for it 🙄 

Thanks for the follow-up !

PR #192, opened 2026-08-30
fix(video): pace UDP shards to reduce burst loss

brongan:
## Problem

Moonshine sends the UDP shards for each encoded frame back-to-back. At higher bitrates, that burst can overflow a queue along the path even when the average stream bitrate fits comfortably.

This remains possible after #157: that change splits oversized GSO submissions into valid datagrams, but the resulting packets are still transmitted as a burst.

The original observation was made with GSO disabled. In two 1440p60, 60 Mbps, 1392-byte packet runs over wired Ethernet, unpaced per-shard sends lost 1.04% and 0.79% of frames. Adding a 100 microsecond gap reduced network frame loss to 0.00% in both runs. A five-minute run with a 125 microsecond gap reported 0.00% network loss and 0.01% jitter loss.

## Implementation

`Pacer` derives its target wire rate from the bitrate requested by the client and a configurable headroom percentage:

```toml
[stream.video]
pacing_bitrate_headroom_percent = 200
pacing_frame_budget_percent = 80
```

Pacing defaults to enabled with 200% bitrate headroom. Setting `pacing_bitrate_headroom_percent` to zero disables it.

The pacer uses Linux `timerfd` for sub-millisecond deadlines. GSO remains enabled: each send contains approximately 500 microseconds of segmented wire data, bounded by the kernel segment and datagram limits. Unusually large frames are accelerated to fit the configured fraction of one frame interval.

The sender reports queue age, send duration, send failures, and missed pacing deadlines every five seconds. It does not discard late encoded frames because later inter-frames may reference them; missed deadlines remain visible in the sender counters.

## Follow-up measurements

These measurements use commit `c744dd484c0a`, a controlled pinwheel source, a Steam Deck connected to an LG TV, routed wired Ethernet, 60 FPS, 60 Mbps, 1392-byte packets, 20% FEC, Moonlight frame pacing enabled, and V-sync disabled. The 4K rows are SDR because the synthetic source did not cause Moonlight to create an HDR10 swapchain.

| GSO | Server pacing | Profile | Network drop | Client jitter drop | Network latency | Decode / render | Host maximum send |
| --- | --- | --- | ---: | ---: | ---: | ---: | ---: |
| enabled | disabled | 1440p60 H.264 | 0.00% | 0.07% | 1 ms | 0.57 / 2.96 ms | 96 us |
| enabled | 200% / 80% | 1440p60 H.264 | 0.00% | 0.00% | 1 ms | 0.60 / 3.19 ms | 10.386 ms |
| enabled | disabled | 4K60 AV1 | 0.00% | 0.03% | 1 ms | 0.53 / 8.37 ms | 89 us |
| enabled | 200% / 80% | 4K60 AV1 | 0.00% | 0.07% | 1 ms | 0.46 / 8.41 ms | 10.414 ms |
| disabled | disabled | 1440p60 H.264 | 0.00% | 0.03% | 1 ms | 0.51 / 2.94 ms | 782 us |
| disabled | 200% / 80% | 1440p60 H.264 | 0.00% | 0.00% | 1 ms | 0.62 / 3.14 ms | 10.681 ms |

Two additional 100-second 1440p60 runs per GSO-enabled setting also reported 0.00% network loss. Both unpaced runs reported 0.04% jitter loss; both paced runs reported 0.00%.

Sender telemetry reported zero send errors, failed GSO chunks, pacing-budget misses, and frame-deadline misses in every row. Deck `UdpRcvbufErrors`, `UdpInErrors`, and `IpInDiscards` did not increase, and each packet capture reported zero kernel drops.

Using consecutive packets separated by at most 100 microseconds as a microburst, pacing reduced the largest observed 1440p burst from 109 packets to 30 and the largest 4K burst from 109 packets to 29. On this path that did not reduce network loss, which was already zero, while completing each frame's sends about 10.3 ms later.

The follow-up data therefore confirms that the pacer reshapes bursts, but does not yet justify enabling that latency trade-off on a working GSO path. The PR remains a draft while deciding whether pacing should be opt-in or limited to the per-shard fallback path.

### DIFF
--- a/moonshine-core/src/session/stream/video/gso_socket.rs
+++ b/moonshine-core/src/session/stream/video/gso_socket.rs
@@ -1,6 +1,7 @@
 use quinn_udp::{Transmit, UdpSockRef, UdpSocketState};
 use tokio::net::UdpSocket;
 
+use super::pacer::Pacer;
 use super::shard_batch::ShardBatch;
 
 /// Maximum payload of one UDP datagram (65535 minus IPv4/UDP headers).
@@ -8,21 +9,6 @@ use super::shard_batch::ShardBatch;
 /// with EMSGSIZE before segmenting. IPv6 allows 20 more bytes; the IPv4
 /// value is safe for both.
 const MAX_UDP_PAYLOAD: usize = 65507;
-
-/// Approximate bytes added to each UDP payload on a VLAN Ethernet path:
-/// IPv4 + UDP + Ethernet/VLAN/FCS + preamble/SFD + inter-packet gap.
-const WIRE_OVERHEAD_BYTES: usize = 70;
-
-#[derive(Clone, Copy, Debug)]
-enum PacingMode {
-	Disabled,
-	Fixed(std::time::Duration),
-	Adaptive {
-		wire_rate_bps: u64,
-		frame_interval: std::time::Duration,
-		frame_budget_percent: u8,
-	},
-}
 
 #[derive(Debug, Default)]
 pub(crate) struct SendBatchResult {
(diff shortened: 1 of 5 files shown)


hgaiser (commented):
Interesting! Never knew about this but it makes sense.

Some notes:

1. Currently the pacing code is a bit scattered throughout. It might be nicer to have a dedicated Pacer struct to handle when shards should be sent. It also makes it easier to test this mechanic.
2. Instead of a fixed pacing_wire_rate_mbps, the pacing ceiling could be derived from the clients requested bitrate with a "headroom" factor.
3. Your implementation sleeps using tokio::time::sleep_until, but according to tokio's documentation: Sleep operates at millisecond granularity and should not be used for tasks that require high-resolution timers..
4. I think we can keep GSO enabled, even when pacing is in effect.
5. I'd be okay with defaulting pacing to on, given the above remarks are resolved.

I'm not sure if we should also add frame dropping if they would be sent after a certain deadline, due to frame pacing; or communicating budget to the encoder to ensure IDR frames are not too large.

PR #193, opened 2026-08-30
fix(xwm): defer X11 resizing until fullscreen

### LINKED ISSUE #188: [Bug] Resonance: A Plague Tale Legacy stuck on black screen on startup
Game just black-screens on startup. No sound, nothing.
Cannot attach any relevant logs so far, gathering more info right now.

---
Host: NVidia 5060 Ti headless machine, latest moonshine-git.
Client: SteamMachine, latest moonlight-qt.

lutyjj:
Fixes #188.

`map_window_request()` resizes every X11 window to the output size before it asks for fullscreen. With Resonance, Wine's toplevel becomes 3840x2160 while the actual rendering child ends up at 3834x2128. That mismatch trips `can_bypass_xwayland()`, the bypass gets rejected, and the stream stays black.

The fix leaves the window geometry alone at map time and lets the existing `enforce_fullscreen_geometry()` path resize it once it is actually fullscreen.

Also drops the `has_game_id()` check. Steam Big Picture requests fullscreen while its app ID is still 0, so keeping that check leaves Steam at 1280x800 instead of filling the output.

Tested end to end at 4K: both Steam login and Big Picture still work correctly and fill in the screen, and the faulty game (A Plague Tale Legacy: Resonance) launches at the correct size too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2451,6 +2451,7 @@ dependencies = [
  "walkdir",
  "wayland-scanner",
  "which 8.0.5",
+ "x11rb",
  "x509-parser",
  "xcursor",
  "zbus",
--- a/moonshine-core/Cargo.toml
+++ b/moonshine-core/Cargo.toml
@@ -82,3 +82,4 @@ zvariant = "5.12.0"
 
 [dev-dependencies]
 tempfile = "3.27.0"
+x11rb = "0.13.2"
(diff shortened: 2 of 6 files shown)


lutyjj:
Putting this into draft for now. It changes map-time geometry handling for every X11 window and makes fullscreen sizing independent of game ID, so I want to test a few more windowed and fullscreen applications first.

PR #197, opened 2026-09-03
Moonlight web only expects positive IDs for applications

BigDiaB:
Moonlight web throws an "invalid digit"-Error when parsing the applist, since "-" is not valid when only looking for u32/positive numbers.

I don't know what the standard practice is for this case, but keeping Moonshine as compatible as possible seems like the right thing in my eyes.


hgaiser:
Bit silly that it works like this, but thanks!

PR #200, opened 2026-09-06
feat(compositor): support native Wayland menus and popups

cfcosta:
> [!NOTE]
> I have implemented this PR using GPT-6 Astra for myself, as Eden (Switch Emulator) would not work without this, basically you need to change inputs at least the first time when running it, and that only happens using the menu and another pop-up window for controller settings.
> I have manually tested it, and the implementation itself does some testing, but feel free to review or close the PR if you prefer.

Implement native `xdg_popup` support in Moonshine’s embedded compositor.
Popup creation, grabs and repositioning were previously unimplemented, preventing native Wayland applications from displaying and interacting with menus, submenus, tooltips and combo boxes correctly.

This change:
- Configures and tracks popup trees, constrains placement to the output, and handles repositioning and unmapping.
- Composes popup surfaces and reports presentation based on visibility.
- Routes keyboard, pointer and touch input, authorizing grabs against input events delivered to the requesting client.
- Preserves touch ownership and restores focus during teardown, rejects stale grabs, and cleans up disconnected clients.
- Adds protocol and rendering regression tests, plus a real-session smoke-test checklist.

Validation passed using the pinned Nix development environment:
- `cargo test --workspace --all-features --locked`: 173 tests passed.
- `cargo clippy --workspace --all-features --all-targets --locked -- -D warnings`
- `cargo fmt --all -- --check`


hgaiser:
Could you give an example of an application that doesn't work regarding menus and popups? I would like to investigate the issue a bit further before committing to a fix.

As an aside, it seems like a large portion of this PR is a test harness and tests that were likely created by your agent so that it could test and verify if a fix worked. For merging this PR it makes more sense if the tests and harness are removed.

cfcosta:
@hgaiser yes, any QT/GTK applications that has a menu should be a fine test target. I tested both Dolphin emulator and Eden emulator.

Ok, I'll be removing the tests and harness from the PR.

desmvu:
FYI, I rebased this onto current `main` (it had gone stale against the focus-selection rework in #212/#214) and dropped the test harness per @hgaiser's comment above: #218

Tested it fixes a real case for me too — RPCS3's Pad Settings controller-type dropdown was invisible (but still eating clicks) before this, renders correctly after.

PR #205, opened 2026-09-09
perf(app_scanner): index icon directories once for fast boxart resolution

### LINKED ISSUE #204: Performance: resolve_missing_boxart traverses entire icon tree per missing boxart (O(N x M) startup freeze)
### Summary

When Moonshine starts up with multiple applications configured that lack an explicit `boxart` field, it enters an un-indexed icon resolution loop that executes a full recursive filesystem walk of all XDG icon directories (`/usr/share/icons`, `/usr/share/pixmaps`, etc.) **once for each missing boxart**.

On a system with extensive icon themes (~90,000–230,000 files in `/usr/share/icons`), having 70+ applications without boxart results in over 17 million directory entries traversed synchronously on the main thread before the server binds to ports.

This pegged a CPU core at 100% and froze startup for **2 minutes and 45 seconds** with no log output, making Moonshine appear completely frozen or failing to start.

---

### Steps to Reproduce

1. Add 50–100 applications to `~/.config/moonshine/config.toml` without specifying `boxart` for each entry (e.g. custom emulator games or ROMs).
2. Start or restart `moonshine` (or `moonshine@.service`).
3. Observe that Moonshine prints the initial startup banner and then completely hangs before printing `Moonshine is ready and waiting for connections.` while consuming 100% CPU.

---

### Root Cause Analysis

1. In `src/main.rs`:

   ```rust
   config.applications.extend(scanned_applications);
   app_scanner::resolve_missing_boxart(&mut config.applications);
   ```

2. In `moonshine-core/src/app_scanner/mod.rs`:

   ```rust
   pub fn resolve_missing_boxart(applications: &mut [ApplicationConfig]) {
       let resolver = desktop::IconResolver::new(true);
       for app in applications.iter_mut() {
           if app.boxart.is_none() {
               app.boxart = resolver.find_icon_by_name(&app.title.to_ascii_lowercase());
           }
       }
   }
   ```

3. In `moonshine-core/src/app_scanner/desktop.rs`:

   `IconResolver::find_icon_by_name()` does not use an index. For every invocation, it iterates through all `self.search_roots` and runs:

   ```rust
   for entry in WalkDir::new(root).follow_links(true) { ... }
   ```

   If N applications lack boxart and there are M files across icon roots, this is an O(N × M) filesystem traversal.

---

### Proposed Fix

Index `icon_search_roots()` once during `IconResolver::new()` into an in-memory `HashMap<String, Vec<IconCandidate>>` keyed by lowercase file stem. Subsequent lookups in `find_icon_by_name()` become instantaneous O(1) hash map lookups.

#### Benchmark Comparison

- **Before:** 74 missing boxarts ≈ 165 seconds startup delay (100% CPU)
- **After (single-pass index):** 74 missing boxarts ≈ 2.4 seconds total startup time (including Vulkan probes and D-Bus init; icon indexing alone took ~0.14s)

mozvip:
Fixes #204

### Problem

`resolve_missing_boxart()` constructs one `IconResolver` and then calls `find_icon_by_name()` once per application missing a `boxart`. Each of those calls performed a full recursive `WalkDir` over every XDG icon search root, so the cost was O(N × M) for N applications and M files under the icon roots.

On a machine with large icon themes and 74 applications without boxart, this froze startup for ~2m45s at 100% CPU with no log output before the server bound its ports.

### Change

`IconResolver::new()` now walks the icon search roots exactly once and builds an in-memory index: a `HashMap` keyed by lowercase file stem mapping to the candidate icon files. `find_icon_by_name()` becomes an O(1) hash lookup against that index instead of a filesystem traversal.

Behaviour is unchanged — the same candidate set is considered and the same preference ordering applies; only the number of times the tree is walked changes (N times to once).

### Benchmark

| | Startup with 74 missing boxarts |
|---|---|
| Before | ~165 s (100% CPU, no output) |
| After | ~2.4 s total, of which ~0.14 s is icon indexing |

(The 2.4s figure is whole-process startup, including Vulkan probes and D-Bus init.)

### Notes

Single commit, touches only `moonshine-core/src/app_scanner/desktop.rs`.

CI's Clippy check currently fails on `clippy::collapsible_if` in `session/compositor/handlers.rs`. That is pre-existing and unrelated to this change (introduced in 4a28808, already on `main`), so it is deliberately left alone here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/moonshine-core/src/app_scanner/desktop.rs
+++ b/moonshine-core/src/app_scanner/desktop.rs
@@ -459,18 +459,58 @@ fn is_executable(path: &Path) -> bool {
 	}
 }
 
+#[derive(Clone)]
+struct IconCandidate {
+	path: PathBuf,
+	extension: Option<String>,
+	score: i32,
+}
+
 pub(super) struct IconResolver {
 	enabled: bool,
 	cache: HashMap<String, Option<PathBuf>>,
-	search_roots: Vec<PathBuf>,
+	index: HashMap<String, Vec<IconCandidate>>,
 }
 
 impl IconResolver {
 	pub(super) fn new(enabled: bool) -> Self {
+		let mut index: HashMap<String, Vec<IconCandidate>> = HashMap::new();
+		if enabled {
+			for root in icon_search_roots() {
+				if !root.exists() {
+					continue;
+				}
+
+				for entry in WalkDir::new(root)
+					.follow_links(true)
+					.into_iter()
+					.filter_map(|entry| entry.ok())
+					.filter(|entry| entry.file_type().is_file())
+				{
+					let path = entry.into_path();
+					if !is_supported_image(&path) {
+						continue;
+					}
+
+					let Some(file_stem) = path.file_stem().and_then(|stem| stem.to_str()) else {
+						continue;
+					};
+
+					let extension = path.extension().and_then(|ext| ext.to_str()).map(|s| s.to_string());
+					let score = score_icon_path(&path);
+
+					index
+						.entry(file_stem.to_ascii_lowercase())
+						.or_default()
+						.push(IconCandidate { path, extension, score });
+				}
+			}
+		}
+
 		Self {
 			enabled,
 			cache: HashMap::new(),
-			search_roots: icon_search_roots(),
+			index,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


hgaiser:
Looks good to me, thanks!

PR #206, opened 2026-09-10
feat(desktop): stream the current KDE Plasma desktop via the XDG portal

brankosimic:
## Summary

Adds **desktop streaming** — streaming the user's current KDE Plasma Wayland desktop (with remote input and system audio) via the XDG desktop portal, like Sunshine — alongside the existing isolated-compositor game sessions. Also fixes a Plasma session startup failure and hardens the video pipeline for captured (non-compositor) frames.

### Desktop streaming (new feature)
- Capture the current desktop via the XDG desktop portal (ScreenCast + RemoteDesktop over PipeWire), with a DMA-BUF fast path and memfd CPU-upload fallback
- Opt-in per host: `[stream] desktop = "Auto" | "On" | "Off"` (default `Auto` — only exposes the `Desktop` app entry when running inside KDE Plasma Wayland with the capture portal available)
- System audio capture (the host's monitor, not a private PulseAudio sink) with a dedicated test suite
- Keyboard/mouse/touch input forwarded back to the session via the RemoteDesktop portal
- Event-driven capture (wakes on PipeWire buffer events, no polling), GPU scaling to the negotiated client resolution, aspect-fit letterboxing with matching input coordinate mapping

### Video pipeline fixes (shared with game sessions)
- Use a plain descriptor set in the pixelforge color converter instead of `VK_EXT_descriptor_buffer` — on RADV/AMD the descriptor-buffer path made `texelFetch` return only the R channel (green-screen stream). *(Patches pixelforge locally until a fix lands upstream — see note below.)*
- Scaler destination uses `SAMPLED` usage; images imported from unimportable DMA-BUFs fall back to CPU upload instead of failing
- GPU-copy images are reported as `GENERAL` layout, not `UNDEFINED`
- Fix black screen when capturing at native monitor resolution

### Plasma session fix
- Use `fail` instead of `replace` as the systemd job mode when starting Plasma sessions, avoiding the `TransactionIsDestructive` conflict where Plasma's systemd integration tears down `graphical-session.target` mid-launch

### Packaging/CI
- Declare `libpipewire` (deb/rpm) and `pipewire`/`glslc` (nix) dependencies, and `libpipewire-dev`/`glslc` in the CI dependency action

## Notes for the maintainer

- The pixelforge patch (`.pixelforge-patch/` + `[patch]` in Cargo.toml) is a vendored copy of pixelforge v0.9.1 with the converter switched to a plain descriptor set. Upstream still uses `VK_EXT_descriptor_buffer`; I can open a separate PR against hgaiser/pixelforge if preferred, and this vendored patch dropped once released.
- Requires testing on AMD/RADV where the descriptor-buffer bug manifests; NVIDIA/Intel likely work with either path.

## Testing

- `cargo clippy --workspace --all-targets` clean, `cargo fmt --check` clean
- Full test suite passes (160 tests incl. new PipeWire format/buffer/audio-capture tests)
- Manually tested streaming a KDE Plasma Wayland desktop from a Moonlight client on RADV/AMD

🤖 I appreciate that this is a large PR — happy to split it into pieces (e.g. portal capture / video fixes / systemd fix) if that's easier to review.


hgaiser:
Oh wow, okay. Thank you for the effort, but with such a big and impactful PR it's better to first reach out to discuss if it's a desirable thing and how it should be made.

If you look back in the git history of moonshine you'd find a period where I was using exactly this (capture with xdg-desktop-portal and pipewire). I made a deliberate choice to move away from that to get more control over the compositor and to avoid requiring a running DE for moonshine to work.

If people want to stream a full desktop, you **can** [run a Wayland compositor](https://github.com/hgaiser/moonshine/blob/main/TIPS.md#run-a-desktop-environment-for-a-full-remote-desktop) (like KDE Plasma) as an application inside Moonshines compositor. I did not test this use-case much and there are likely issues with this approach, but I would much rather fix those issues and avoid adding a secondary capture path.

That said, I think it makes more sense to use something like [RustDesk](https://rustdesk.com/) to use full desktop streaming, but maybe that's just me.

So to summarize, thank you again for the effort, but I am not going to merge this for the above reasons.

Issue #208, opened 2026-09-10
Keyboard handling broken due to steam overlay fix (4a28808)


KarlVogel:
Commit 4a28808 fix: improve steam integration, fixing steams overlay 

This seems to have broken some stuff. Zomboid & Factorio (both linux native games) have non working keyboard with it, also pressing ESC terminates the game. Proton-based game like "AI Limit" also has non functional keyboard (ESC doesnt exit that game though).

Also noticed steam big screen mode doesnt have an "Exit Steam" option with that patch in, it only shows an "Exit to desktop".

hgaiser:
Ah shoot.

Should be fixed with https://github.com/hgaiser/moonshine/commit/2b122d7b2805104313b1c766f5f4cc3bb7e8cdb0, can you test on your end?

baberabb:
I think I have the same issue with Baldur's Gate 3 (also linux native, I believe), not recognising controller inputs on `v0.16`.  I stream from a Fedora host to my steamdeck, and the game never picks up on the controller. Inputs stays with the Steam client in the background (can hear the UI sound effects). The game also always crashes shortly after a couple of random button presses. Downgrading to `0.15` fixes this.

I had claude write up some more details, in case it's helpful:

<details>
<summary>claude sumary</summary>

### Summary

Since 0.16.0 started impersonating gamescope, Steam Big Picture keeps ownership of the
gamepad for the entire session. The game window is created, mapped and correctly tagged
with `STEAM_GAME`, but moonshine never adds the game's appid to `GAMESCOPE_FOCUSABLE_APPS`
and never moves `GAMESCOPE_FOCUSED_APP` / `GAMESCOPE_FOCUSED_WINDOW` off the Steam Big
Picture window.

Because Steam now believes it is running under gamescope, it trusts those atoms to decide
who owns controller input. They say the Steam UI is focused, so Steam consumes every
gamepad event itself: button presses navigate Big Picture invisibly behind the running
game, and eventually one of them lands on "stop game" and kicks you back to the Steam UI.

0.15.0 is not affected.

### Environment

| | |
|---|---|
| moonshine | 0.16.0-1.x86_64 (official release RPM) |
| last known good | 0.15.0-1.x86_64 |
| OS | Fedora 44, fc44 |
| Steam | steam-1.0.0.87-1.fc44 (RPM Fusion, native — not Flatpak) |
| Game | Baldur's Gate 3 (appid 1086940), Proton, X11 via Xwayland |
[...]

PR #212, opened 2026-09-12
 fix: re-evaluate focus when Steam updates `GAMESCOPECTRL_BASELAYER_APPID`

baberabb:
Had Claude put together a fix for #208. Tested on my steamdeck on BG3 and Kingdom Come Deliverance. Controller inputs are now getting passed through successfully.

I can't follow the code well enough to review it properly so feel free to use this as a starting point, and/or close it if it's not the right approach.

### The bug

Since moonshine started impersonating gamescope, Steam Big Picture keeps ownership of the
gamepad for the whole session. The game launches and renders normally, but every controller
press goes to the Steam UI behind it — navigating Big Picture invisibly until something lands
on "stop game" and drops you back to the UI. Touch input is unaffected, so the game looks
responsive to touch and completely deaf to the pad.

### Root cause

Steam hands focus to a launched game by rewriting `GAMESCOPECTRL_BASELAYER_APPID` on the
**root window**, moving the game's app ID ahead of its own `769` once the game is up.

That property lives on the root window, which is not a window Smithay's XWM manages, so its
changes never reach `XwmHandler::property_notify`. Nothing else re-runs focus selection in
steady state — once a game is running and no window events are firing, there is no trigger.
So the handoff is read exactly once, while the Steam UI is still legitimately in front, and
focus stays there for the rest of the session.

The observable symptom, with the game mapped and in the foreground on its own Xwayland
server (`:3`):

```
GAMESCOPE_FOCUS_DISPLAY(UTF8_STRING)       = ":3"                      # correct
GAMESCOPECTRL_BASELAYER_APPID(CARDINAL)    = 413091, 769, 1086940      # Steam published it
GAMESCOPE_FOCUSED_WINDOW(CARDINAL)         = 37748793                  # Steam Big Picture
GAMESCOPE_FOCUSED_APP(CARDINAL)            = 769                       # Steam UI
GAMESCOPE_FOCUSABLE_APPS(CARDINAL)         = 769                       # game never a candidate
```

Both windows are present on that server and the game window is tagged correctly:

```
0x2400039  WM_CLASS "steamwebhelper"  STEAM_GAME 769
0x3000002  WM_CLASS "bg3"             STEAM_GAME 1086940
```

So the app ID is known and published — it is simply never read again after the initial
evaluation.

### The fix

Subscribe to `PropertyNotify` on the root window and re-evaluate focus when Steam's focus
control changes.

- `X11Focus::watch_focus_control()` selects `PropertyChangeMask` on the root via the existing
  dedicated `X11Focus` connection. `PropertyChangeMask` is not an exclusive mask (unlike
  `SubstructureRedirectMask`), so this does not disturb the XWM's own connection to the same
  server.
- `X11Focus::connection_fd()` exposes the connection fd, registered with calloop as a
  level-triggered `Generic` READ source.
- `X11Focus::drain_focus_control_change()` drains the queue on wake and reports whether a
  `PropertyNotify` for `GAMESCOPECTRL_BASELAYER_APPID` or `GAMESCOPECTRL_BASELAYER_WINDOW`
  arrived on the root. The queue is always drained fully, even after a match, so nothing is
  left behind to immediately re-wake the loop.
- On a match, `reevaluate_focus()` runs and focus moves to the game.

Lifetime: the source is registered right after `X11Focus::open()` in the XWayland-ready path,
and removed in `shutdown_session_processes()` **before** `X11Focus` is dropped and the display
closed, so the borrowed fd is never left registered against a dead connection.

Four Xlib symbols are added (`XSelectInput`, `XPending`, `XNextEvent`, `XConnectionNumber`)
following the existing optional-`dlsym` pattern — if any fail to resolve, the watch is skipped
with a warning and behaviour is unchanged from today.

### Testing

Fedora 44, RTX 5090 (NVIDIA 610.57.04), native Steam, Baldur's Gate 3 under Proton on
Xwayland, Moonlight on a handheld client.

- Before: controller input reaches Big Picture only; `GAMESCOPE_FOCUSED_APP` stays at `769`
  for the entire session.
- After: focus moves to the game when Steam publishes the handoff, and the controller reaches
  the game.
- Verified the symptom is absent on 0.15.0 and present on 0.16.0, same config and same game.

### Notes for review

- `XPropertyEvent` is declared `#[repr(C)]` with all fields so `window` and `atom` land at the
  correct offsets; only those two are read. The event buffer is the full 24-`long` `XEvent`
  union size, as Xlib requires — `XNextEvent` writes the whole union regardless of variant.
- This is X11-only. The 0.16.0 `xdg-activation` change covers Wayland-native clients getting
  keyboard focus; it does not touch the gamescope app-ID atoms Steam uses to route the
  gamepad, so X11 games were left with no way to take focus.


hgaiser:
Thanks! I took some parts, reverted some others, LGTM 👍

Issue #93, opened 2026-05-20
DX11 games having black screen, audio and input still works


desmvu:
I tried streaming Witcher 3 in DirectX 11 mode using Moonshine, audio works and input works but screen is black. I have tried restarting the streaming server, didn't work, tried DX12 mode and it works.

My streaming server:

- AMD Ryzen 7800x3D
- 32gb RAM DDR5
- RTX 3080 10gb
- cachyOS 7.0.9, KDE Plasma
- Moonshine 0.11.0

My streaming client:

- AMD Ryzen 5600x
- 24gb DDR4
- RTX 3060
- cachyOS 7.0.9, KDE Plasma
- Moonlight Flatpak

hgaiser:
Thanks for the report! Luckily I have Witcher 3 so I can test this locally :).

desmvu:
To clarify, DX12 mode now also doesn't work. Same issue, video is black but audio and input works. I think DX12 worked because I was on 0.10.0 and I just coincidentally switched to DX11 when I updated Moonshine to 0.11.0.

mrhexzero:
Can confirm this is also an issue on CEMU

hgaiser:
For me, Witcher 3 works both in DX11 and DX12 mode. Can you try with the latest `moonshine-git` build (or if you are running from source directly, make sure the `moonshine-wsi` layer is installed).

I also tried with moonshine v0.11.0, which also works =\

Note: there was an issue with Witcher 3 not receiving keyboard and mouse input, which should be resolved with https://github.com/hgaiser/moonshine/pull/108

hgaiser:
This bug has been driving me insane. I was testing https://github.com/hgaiser/moonshine/pull/108 and got black screen with Witcher 3. I thought great, then I can debug this issue.

I was debugging it, found some potential causes but nothing worked. I tried "wrapping" Baldurs Gate 3 (which had the same issue) with gamescope by setting launch options. That worked (except for HDR, but that doesn't surprise me that much). I revert back to normal launch options and then that too works 😕 . No matter what I do I can't get it to break now. I tried rebooting, reverting to older versions, clearing caches.

This aligns *vaguely* with what you were seeing @desmvu , where you thought DX12 worked but then all of a sudden it too didn't work.

There's a really interesting but annoying bug here.. it could be some race condition that for some reason doesn't trigger for me anymore, but I can't really debug it further now. Plus it has drained me of my sanity, so I need to look at something else for a bit 🥲

[...]

PR #214, opened 2026-09-13
fix(compositor): native wayland focus, HDR and window stacking fixes

urwrstkn8mare:
some minor fixes here and there related to the compositor.

combined with your recent changes for #208 (there was a small issue with the window ordering), this fixed the issues I was having, which looked just like the ones described in #208.

it also fixes:
- HDR not working at all for me
- some wayland games getting a window border drawn around them, caused by another PR adding a maximised state to windows
- the black screen issue from #93

important to note I've only tested this myself on a 9070 xt system, so it'd be great if others could check it works for them too. (used LLMs to help track these down)


moonshine-core/src/session/compositor/color_management.rs
```diff
@@ -246,7 +246,7 @@ impl ColorManagementState {
 	/// Create a new state and register the protocol globals.
 	pub fn new(display: &DisplayHandle, hdr: bool) -> Self {
 		// Advertise wp_color_manager_v1 (protocol version 1).
```
KarlVogel: comment needs an update 1->3

urwrstkn8mare: fixed in 82df21e, thanks

hgaiser:
Thanks for helping, looks good to me!

PR #218, opened 2026-09-17
feat(compositor): support native Wayland menus and popups (rebase of #200)

desmvu:
## Summary

Rebases @cfcosta's native `xdg_popup` support from #200 onto current `main`. `new_popup`/`grab`/`reposition_request` were complete no-ops, so any Wayland-native app's dropdowns/context menus/combo boxes (Qt-on-Wayland, GTK-on-Wayland) never rendered — the popup surface mapped but nothing composited it, while pointer input still hit whatever was underneath. Confirmed this against RPCS3's Pad Settings controller-type dropdown, and it matches the Eden/Dolphin repro @cfcosta gave in #200's discussion.

This rebase was needed because #200 predates the gamescope focus-selection port (#212/#214) that shipped in 0.16.1 and had gone stale/unmergeable, and per your review comment there:

> For merging this PR it makes more sense if the tests and harness are removed.

So this branch:
- Drops the ~1050-line AI-generated test harness (`test_state.rs`, `tests.rs`, `popup_*_tests.rs`) entirely.
- Keeps the current `build_candidates`/`space_render_elements_with_override`/`render_output` implementations from the focus-selection port rather than #200's older equivalents, layering popup rendering on top of them instead of replacing them.
- Reverts the `renderer` field from `Option<GlesRenderer>` back to plain `GlesRenderer` — the `Option` wrapping only existed to support the deleted test harness's mock renderer.
- Updates the touch grab code (`popup_touch.rs`, `popup_touch_focus.rs`) to the current pinned Smithay `TouchGrab`/`TouchTarget` trait signatures (no more trailing `Serial` param on most methods; `frame`/`cancel` now take `FrameMarker`). Dropped one edge-case touch-cancellation replay in `PopupTouchGrab::cancel` that depended on constructing a `FrameMarker`, which has no public constructor in this Smithay revision — not required for pointer-driven popups like RPCS3's.

Commit authorship is preserved from #200 (@cfcosta) plus one fixup commit from me for the remaining API drift.

## Test plan

- [x] `cargo build --release --workspace`
- [x] `cargo test --release --workspace` (128 + 17 passing)
- [x] `cargo clippy --release --workspace --all-targets -- -D warnings`
- [x] Manually verified: streamed RPCS3 via Moonshine, opened Pad Settings, controller-type dropdown now renders and is clickable (previously invisible but still eating clicks).


hgaiser:
Hey, thanks for looking into this, but I'm working on a big rewrite of the compositor to make it easier to do standalone (read: without moonshine) testing. Any work on the current compositor will fail to merge after that work is done.

I prefer to wait until that rewrite is finished to revisit these PRs.

